PR #12, opened 2024-08-21
Add abstract dropdown page

MaHaWo:
- adds a simple dropdown page with the actual content being passed as arguments. 
  - add component for selection of categorical data
  - add component that eats an array of data objects 
- creates a dummydropdown component to see how it works. 
- adds routes from frontpage to dropdown to next dropdown page, just to make it do something for now. 
- adds route for infopage (temporary).


MaHaWo:
- changes since last discussion: 
  - get rid of js class 
  - componentize the data selection a bit better 
  - make the selection element more exchangeable

MaHaWo:
@lkeegan you want to have another look?

lkeegan (commented):
lgtm!

PR #14, opened 2024-08-21
Componentize front page better

MaHaWo:
This is a little refactoring PR: 

- makes smaller components: 
  - bottom navbar items that link somewhere 
  - top bar icons  with tooltip
  - cards with header, summary and button to link 

- put top navbar into outermost `+layout.svelte` because we will have it everywhere. 

eventually, `AbstractContent.svelte` will go away when we have more pages and the link hierarchy becomes apparent, such that elements that are needed everywhere can be put into the +layout.svelte files. The bottom navbar can be used for everything inside a `Bestandsaufnahme` (survey/stock taking) for example, or for the registration process.

### DIFF
--- a/src/lib/components/AbstractContent.svelte
+++ b/src/lib/components/AbstractContent.svelte
@@ -7,14 +7,21 @@
 		NavBrand,
 		NavHamburger,
 		NavLi,
-		NavUl,
-		Tooltip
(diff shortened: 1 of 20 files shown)


lkeegan (approved):
lgtm!

PR #16, opened 2024-08-22
add tests workflow file

MaHaWo:
- add a workflow file for unit tests (and others later on)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,13 +15,17 @@ jobs:
         - name: Checkout code
           uses: actions/checkout@v4
 
-        - name: Set up environment 
-          uses: actions/setup-node@v4
-          with: 
-            node-version: '20'
+        - uses: pnpm/action-setup@v4
+          with:
+            version: 9
+
+        - uses: actions/setup-node@v4
+          with:
+            node-version: 22
+            cache: 'pnpm'
 
         - name: Install dependencies
-          run: npm install
+          run: pnpm install
 
         - name: Run tests
-          run: npx vitest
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,27 @@
...
+
+jobs:
+    unit-tests: 
+      runs-on: ubuntu-latest
+
+      steps: 
+        - name: Checkout code
+          uses: actions/checkout@v4
+
+        - name: Set up environment
```
lkeegan, line 18: pnpm is a faster replacement for npm, see e.g. this workflow: https://github.com/spatial-model-editor/spatial-model-editor.github.io/blob/3d93eac74e8f2b0e1b29429cb5c849602bb19093/.github/workflows/deploy.yml#L11-L20

.github/workflows/ci.yml
```diff
@@ -0,0 +1,27 @@
...
+        - name: Set up environment 
+          uses: actions/setup-node@v4
+          with: 
+            node-version: '20'
+
+        - name: Install dependencies
+          run: npm install
+
+        - name: Run tests
+          run: npx vitest
```
lkeegan, line 27: this is already a command in package.json, so we can just do `npm test` (or `pnpm test`)

.github/workflows/ci.yml
MaHaWo: :+1:

.github/workflows/ci.yml
MaHaWo: :+1:

PR #17, opened 2024-08-22
Create abstract table display

### LINKED ISSUE #4: Create abstract table overview page
A number of things in mondey are presented in the form of tables with color coded cells: 
- feedback page 
- regular survey completion status
- children overview 
- add tooltips on hover.  

A component that abstracts this away would be good. I'm not yet sure if it's possible or a good idea to throw the children overview here, too, because the [gallery flowbite template](https://flowbite-svelte.com/docs/components/gallery)  than the [flowbite table](https://flowbite-svelte.com/docs/components/table).

MaHaWo:
- creates a searchable table component with cells optionally being colored to indicate status
- create components for header and body of the table 
- add some unit tests for internal javascript logic using svelte-testing
- fix a few minor errors 
- added only minimal styling since I'm not yet sure how the whole thing will turn out when it is bound together.

resolves #4


lkeegan (approved):
looks nice! The filtering might be useful for admin views later. I guess a simple feature we still need here is for users to be able to click on one of the cells of this table to continue filling in that survey

Issue #10, opened 2024-08-20
Dummy backend


MaHaWo:
perhaps we should have a dummy backend based on svelte stores or so with which to imitate backend requests?

MaHaWo:
This has largely been implemented now, but at the current stage it needs more work to be turned into the frontend API that talks to the backend.

PR #22, opened 2024-08-27
Add stores for data

MaHaWo:
*This is all somewhat tentative since I'm not proficient with web data management yet*

#### what it does currently
-----------------------------------------

- goal: add svelte stores for children and content which can later be used to fetch and store data from the backend. 
- use svelte-stores to store global data in and 'emulate' backend calls. 
- There is no fixed API for yet for how these calls should be made in the end. Currently there's just some `add` , `remove` and `fetch` functions that operate directly on the prepopulated stores that contains only dummy data. For this reason, this PR doesn´t close the #10  yet. 
  - 2 stores: children (contains userkey -> array of children objects, which are themselves dictionaries), content (contains data: surveys -> objects). 
- currently there are no subscriptions because the stores do not change values at runtime yet.
- move dummy data defined in the components into the stores and use via the respective stores then. 
-  create a dynamic routing for children and surveys: Children summary page and survey pages are created dynamically depending on the data in the stores. This currently uses `load` to pass data from page to page --> *is there something better or is this acceptable?*
- add tests for functionality of stores where deemed relevant.
- use typescript in the `xStore` and `xStore.test` files (will slowly migate `<script>` blocks and `.js` files to typescript as things get refactored).
- there is no user store yet. This is because the user data management system is the least clear to me and the current login/register pages need deeper refactoring. For this reason, the child registration page does not make use of the childrenstore either. 
- error handling is missing still. 

Sorry this got so ugly, I went back and forth a lot on it...


MaHaWo:
@lkeegan Is there something to do here still?

lkeegan (approved):
lgtm

PR #28, opened 2024-08-30
Make children registration work

MaHaWo:
- New children can be registered now for a hardcoded 'dummyuser' and appear on the overview page 
 -  build everything needed in the `ChildrenDataInput` component from flowbite and svelte components and do not use the `AbstractRegistration` component for now. 
 - add some basic input validation via button events. 
 - make use of localstorage for now because we don´t have a database yet to have persistent data across pages in the childrenstore 
 - have better separation of form data and input data.
 - add `AlertMessage` component that appears at the top of the page body and can indicate errors or things that one needs to pay attention to. 
- Not done here: overview page for children and their data, which on demand can be edited.


lkeegan (approved):
looks good! some minor ui suggestions:

- dual validate / submit button is a bit unclear to me
  - a simple submit button would I think be clearer (which calls validate & either submits or not)
- when a submission is invalid
  - would be nice for the page to move up to the first invalid field & set focus there as well as the existing highlighting & message

PR #32, opened 2024-08-30
consolidate stores into classes and add userstore

MaHaWo:
- add basicStore 
- use it to implement children, user and contentstores
- make them all typescript classes (?)
- use singleton pattern to make sure there's only one instance 
- adjust tests
- FIX: make some imports generic that previously were not


lkeegan (approved):
lgtm!

Issue #26, opened 2024-08-28
add user store


MaHaWo:
- once the user registration is reworked such that functionality is included we can add the user store analogously to the children store. 
- needs to be included into the user registration and login page.

PR #34, opened 2024-09-02
Add functionality to login page

MaHaWo:
### Depends on #26 !

- make login page functional 
  - [x] add 'loggedIn' element to userStore which is null when nobody is logged in and holds the user's ID otherwise 
  - [x] add UserProfile page that shows up when the avatar icon is clicked and shows either user info and buttons to go further if logged in or login/register button if not 
  - [x] add validation function that stores the logged in userID and adjusts error messages etc
  - [x] add `Input` component that handles internally the coloring of fields etc. Didn't quite manage to focus this however when it's field is missing, because it is a component and not a DOM object. solution could be to ditch flowbite-svelte in this small stuff and ship our own which then uses `<input></input>` internally? Maybe I'm just missing something... 
  
**tbd separately because generic for all form elements in the project so far** 
- make keyboard navigation and focus work the way it is usually expected: 
  - automatic jump to next element when on is filled and enter is hit 
  - hit on 'enter' will submit the form and act like a click on login 
  - tooltips for elements 
  - the loading and writing of stores to localstorage is messy and should imho be more concisely.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,7 @@ vite.config.ts.timestamp-*
 
 # packaging
 package-lock.json
+
+
+# extension data 
+.VSCodeCounter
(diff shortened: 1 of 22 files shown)


lkeegan (approved):
"remember me" checkbox doesn't seem to work? otherwise lgtm!

MaHaWo:
it should work after the last change.

PR #35, opened 2024-09-04
Add Milestone component

lkeegan:
- corresponds to slides 14 & 15 from MONDEY-Neu-Design_28-08-24
- use Breadcrumb for backwards navigation instead of multiple "zurück" links
- each answer button has a different shade of green as background colour
- mouseover tooltips for answer buttons match colour of answer button
- vertical layout on small screens, horizontal on large screens
- checkbox to automatically go to the next question without having to click weiter
- förderhilfen is in an Accordian - initially hidden
- add temporary /milestone route to demo the component
  - add temporary baby images for demo


lkeegan:
image

MaHaWo:
lgtm. 

only minor thing: We don´t want people to be able to go back to the last milestone? (no 'zurück') button

MaHaWo (approved):
lgtm. 
Minor things: 
- We don´t want people to be able to go back to earlier milestones?
- At some point I guess we will have to decide on a color scheme, since we have two contrasting ones now.

PR #38, opened 2024-09-05
Add initial i18n support with en/de

### LINKED ISSUE #9: Localization/support for english
This  has to be taken care of at some point in the future.

lkeegan:
- add language choice to top navbar
- for now only the milestone component is available in multiple language
- text for other components can be added to the json files in the src/lib/locales folder
- another language can be added by adding a new json file to the locales folder
- uses the svelte-i18n library
- resolves #9


MaHaWo (approved):
lgtm

PR #57, opened 2024-09-16
Create better input component

MaHaWo:
- add a `DataInput` component that has basic validation, formatting and customizable basic component built in
- form validation by default only checks if they elements are not empty. More logic can be passed via a function. 
- use it in place of `Input` and `Select` custom components
- delete the latter two components
- delete `AbstractDataInput` 
- delete `surveyX` page. This was a placeholder for the milestones and hence is not needed anymore since milestones are actively worked on and available now 
- refactor childrenRegistration data
- event handlers are currently only selectively exposed, because a generic, programmatic ability to pass them to components is only implemented in svelte 5 (see [here](https://github.com/sveltejs/svelte/issues/2837#issuecomment-1848225140))
- allow for `MultiSelect` in the data input for language, nationality, disabilities... 
- there is no automatic type detection. This functionality would need to work in a very general way on arbitrary input, and it's not clear to me that the effort to implement is is worth the gain when the type is static anyway and can be hardcoded for the element in question.
- improve handling of non-identical passwords on user registration


lkeegan (approved):
lgtm!

PR #58, opened 2024-09-16
Display multiple images in Milestone component

### LINKED ISSUE #51: Multiple images in frontend Milestone component as carousel

lkeegan:
- milestone component cycles through available images for that milestone
- transition via opacity every 5secs
- provide list of image urls in demo data instead of a single image
- remove padding on mobile devices to use more of the screen
- resolves #51

### DIFF
--- a/frontend/src/lib/components/Milestone.svelte
+++ b/frontend/src/lib/components/Milestone.svelte
@@ -11,11 +11,17 @@
     let autoGoToNextMilestone: boolean = false;
 
     // build list of possible image urls at build time to be able to dynamically use them at run time:
-    const images: Record<string, string> = import.meta.glob('$lib/assets/*.jpg', {
+    const img_urls: Record<string, string> = import.meta.glob('$lib/assets/*.jpg', {
         eager: true,
         query: '?url',
         import: 'default'
     });
+
+    let currentImageIndex:number = 0;
+    const imageInterval = 5000;
+    setInterval(() => {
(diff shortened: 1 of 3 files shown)


frontend/src/lib/components/Milestone.svelte
```diff
@@ -11,12 +11,18 @@
...
     // build list of possible image urls at build time to be able to dynamically use them at run time:
-    const images: Record<string, string> = import.meta.glob('$lib/assets/*.jpg', {
+    const img_urls: Record<string, string> = import.meta.glob('$lib/assets/*.jpg', {
         eager: true,
         query: '?url',
         import: 'default'
     });
 
+    let currentImageIndex:number = 0;
+    const imageInterval = 5000;
```
MaHaWo, line 21: is 5s perhaps a little long as a delay between images? I'm not sure where  a good delay would be that makes people realize that there is content to look at here before they get impatient and start doing things....

MaHaWo (approved):
lgtm. one minor remark wrt delay between images

frontend/src/lib/components/Milestone.svelte
lkeegan: Yes I'll make an issue for this - it's currently a bit long if you're just clicking an answer, but if you open the dropdown help text and try to read it then (at least for me) I found it a bit annoying when it kept updating - could also just cycle once then stop I guess.

lkeegan: https://github.com/ssciwr/mondey-frontend-prototype/issues/62

PR #60, opened 2024-09-16
Add text input for custom select option

MaHaWo:
- allow for special element `textTrigger` in `DataInput` components that triggers the display of a textarea to fill in values that are not in the list of selectable values. 
- use this in childrenRegistration and userDataInput by having an `others` option which upon selection allows for other values to be added 
- refactors `DataInput` component to allow for these additional values:  
  - have `additionalValue` to bind to textarea input
  - when `value` is equal to `textTrigger` the additional text field is displayed 
  - the textfield has a `on:blur`  eventhandler that updates `value` by replacing the element equal to `textTrigger` in `basicValue` with the `additionalValue`, such that we can have,e .g., language selection like `[deutsch, enlish(uk), mongolian]` even though mongolian isn´t in the list of languages: The user starts by selecting languages and 'other', then the textfield is displayed and the user provides additional input. Once the textfield is left, on:blur kicks in and the the 'other' element in the selected languages is replaced with the user-defined content.
  -  refactor the `userDataInput`  component such that it works in the same way as the `childrenRegistration` component. This mostly pertains to the datastructure.


lkeegan (approved):
lgtm as is - as discussed we'll refactor the data input in a separate issue&pr

Issue #45, opened 2024-09-13
[component architecture] Milestone group component


lkeegan:
We need a milestone group component, that displays all the milestone groups.

This could be integrated visually with the milestone component, keeping the same breadcrumb navigation bar at the top for consistency.

There could be two levels of display, one that lists the individual milestones from a group, e.g. "Grobmotorik"

Then a higher level one that lists all the milestone groups, e.g. "Meilensteine"

PR #67, opened 2024-09-18
Create milestone overview page

MaHaWo:
- add overview over all the different milestones by using `CardDisplay` and `GalleryDisplay` in much the same way it is used for the `ChildrenGallery`.
- add new page for demo display
- corresponds to the 'higher level' mentioned in  #45 
- add optional progress bar to the card display that shows current completion status for the respective milestone questionaire (only dummy values atm)
- optional image 
- refactoring of card styling to acommodate expanded demands. especially for column things. Still has a bug: Heading long => overflows smaller card. However, since a `CardDisplay` can have multiple auxilliary elements now (buttton, progressbar,...) it would make sense to refactor it into something more general in a separate PR. 
- reconsidered stuff from IRL discussion and added breadcrumbs component for navigation through userLand.  
- access: `localhostURL/milestonegroup`


frontend/src/lib/components/MilestoneGroup.svelte
```diff
@@ -0,0 +1,41 @@
+<script lang="ts">
+	import Breadcrumps from '$lib/components/Breadcrumps.svelte';
+	import CardDisplay from '$lib/components/DataDisplay/CardDisplay.svelte';
+	import GalleryDisplay from '$lib/components/DataDisplay/GalleryDisplay.svelte';
+	export let breadcrumpdata: any[] = [];
+	export let milestonedata: any[] = [];
+
+	// FIXME:styling has no business being here... not sure where to put it though given thatparts of it are data dependent
```
lkeegan, line 8: I think the usual solution to this is to make a component, e.g. we could have our own ProgressBar which takes progress as a prop and is responsible for its styling, etc

lkeegan (commented):
looks very nice!

minor comments:

- you could add a link to this in the component demos part of the readme
- does the search button do anything?
- "breadcrump typo" in filenames/source code

frontend/src/lib/components/MilestoneGroup.svelte
MaHaWo: noted. The progressbar will indeed need it's own thing I think because the flowbite-svelte one is rather limited atm

PR #68, opened 2024-09-18
Add overview page for individual milestone

MaHaWo:
- uses much the same visual buildingblocks as the milestonegroup page: `table of cards`
- uses button exposed by `CardDisplay` component to indicate status and allow for access to milestonepage
- allows for searching milestones by status or by part of header 
- displays status bar. Intention is to update it when new milestones are completed, but this has not yet been included
- uses breadcrumbs for navigation.
- adds ability to have a configurable Icon in the `CardDisplay` button 
- adds customizable `searchPlaceHolder` and seach function to the `GalleryDisplay`

### DIFF
--- a/frontend/src/lib/components/DataDisplay/CardDisplay.svelte
+++ b/frontend/src/lib/components/DataDisplay/CardDisplay.svelte
@@ -62,4 +62,11 @@
 			labelInsideClass={styleProps.progress?.labelInsideClass}
 		/>
 	{/if}
+
+	<!-- This will be generalized to replace button and progressbar later  -->
+	{#if data.auxilliary}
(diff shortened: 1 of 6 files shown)


lkeegan (approved):
thanks - looks good, a few suggestions:

- cards should be clickable like the other cards
- we could remove most/all of the heading text and just have the searchbar and the cards
  - user has already chosen this milestone group, they presumably just want to see what the milestones are
- cards could have a background colour + icon to indicate done or not rather than an additional button
- maybe the description text is not required?
  - this would allow many more cards to be visible on the screen at one time
- cards to be done should come first

MaHaWo:
implemented suggestions as discussed

PR #73, opened 2024-09-20
Refactor data input element (again)

MaHaWo:
- add custom `RadioList` component with support for single or multiple value input
- remove weird behavior when having additional values provided through a textbox. the additional stuff lives in a separate variable now.
- integrate `RadioList` into `ChildRegistration` and `UserRegistration` pages. Make things with less than five elements RadioList, data input elements with more I left as dropdowns because I found it visually overwhelming. 
- make `ChildrenRegistration` and `UserRegistration`  the same since they do the same thing 
  Question: Should they be [Modal](https://flowbite-svelte.com/docs/components/modal) perhaps? 
- move all the data definitions into the respective `+page.svelte` from the component svelte files.


MaHaWo:
sry, there's a bunch of automatic reformatting stuff in there.

MaHaWo:
@lkeegan you want to have another look

lkeegan (approved):
lgtm - as discussed I'll try to construct the JSON that the backend will provide, then we can revisit this

PR #87, opened 2024-09-24
Add search function for milestone components

MaHaWo:
- makes `GalleryDisplay` accept an additional variable `searchData` (an array of objects) that defines categories of data to be searched for. 
- add a dropdown button to select the search category. Per default the first is used. 
  - TODO remains: need to focus input field automatically when category is selected, but usual `bind:this={thing}` => `thing.focus()`  didn´t work. Left for another PR that fixes this kind of usability stuff 
-  Implement a number of search categories for `milestoneoverview` and `milestonegroup`. Which ones we want actually can be decided quickly, since we just need to exchange an element in an array to add or remove them.

### DIFF
--- a/frontend/src/routes/milestonegroup/+page.svelte
+++ b/frontend/src/routes/milestonegroup/+page.svelte
@@ -23,14 +23,55 @@
 		}
 	];
 
+	const milestoneData = [
+		{
+			number: 1,
+			title: 'Alleine von Stufe/Absatz springen',
+			desc: 'Kind springt mit beiden Beinen gleichzeitig hoch und überwindet im Sprung freihändig einen (kleinen) Absatz oder eine Stufe. Es kommt sicher wieder im Stand auf.',
+			observation:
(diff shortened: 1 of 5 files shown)


frontend/src/routes/milestonegroup/+page.svelte
```diff
@@ -23,77 +23,232 @@
...
+
+	interface DataElement {
+		header: string;
+		summary: string;
+		milestoneData: any;
+		progress: number;
+		[key: string]: any;
+	}
+
+	function searchByStatus(data: DataElement[], dummy: any, key: string): DataElement[] {
```
lkeegan, line 162: what is `dummy` for?

frontend/src/routes/milestonegroup/+page.svelte
```diff
@@ -23,77 +23,232 @@
...
+		} else {
+			return data.filter((item) => {
+				// button label contains info about completion status => use for search
+				if (key === 'fertig') {
+					return item.progress === 1;
+				} else if (key === 'unfertig') {
+					return item.progress < 1;
+				} else {
+					return item.header.toLowerCase().includes(key.toLowerCase());
+				}
```
lkeegan, line 174: If we want to filter for complete/incomplete it might be nicer to have a couple of checkboxes for this under the search box, versus having the user type fertig

frontend/src/routes/milestonegroup/+page.svelte
```diff
@@ -23,77 +23,232 @@
...
+		} else {
+			return data.filter((item) => {
+				return item.milestoneData.some((element: MilestoneData) =>
+					element.title.toLowerCase().includes(key.toLowerCase())
+				);
+			});
+		}
+	}
+
+	// README: this is slow and quite a bit of work because a lot of text has to be searched. Kill it?
```
lkeegan, line 211: I'd keep for now and revisit this once we have realistic data

lkeegan (approved):
looks nice! a few minor questions/comments

also it would be good if the size of the cards didn't change as the number of items varies, not sure if this is an easy fix or needs to be looked at separately?

frontend/src/routes/milestonegroup/+page.svelte
MaHaWo: needed because the underlying logic in `GalleryDisplay` allows for restriciting search to configurable columns/elements of the displayed gallery and thus needs an additional argument. This is not needed here and therefore uses a dummy value.

frontend/src/routes/milestonegroup/+page.svelte
MaHaWo: hm, yeah.... that would require an additional element in the UI however and it would add significant complexity to it.

frontend/src/routes/milestonegroup/+page.svelte
MaHaWo: ok... I refactored a bit and got rid of it.

PR #91, opened 2024-09-25
Add milestones to backend

lkeegan:
- add/update Milestone models and admin endpoints in backend
- add Milestone components to frontend

Also make endpoint naming more consistent

- use plurals for all collections of entities
- use PUT instead of PATCH since these endpoints take the entire entity, not just selected fields to update


MaHaWo:
... started review, but need some more time ...

MaHaWo (approved):

PR #95, opened 2024-09-27
Bind together userland

MaHaWo:
- have user landing page when they are logged in with two tabs: one for user data one for all the children stuff
- some steps towards single page app:
- add a store for components 
- add possibility for `Breadcrumbs` component to have onclick event & button instead of link
- use on:click event to make dynamic loading of components possible in a single page with breadcrumbs
- make it such that the children components are rendered into the tab on the landing page. Breadcrumbs used for navigation within children area as it was. Most routes are mad obsolete by this, which hence have been deleted. 
- refactor out crappy logic stuff that will be replaced by backend calls into file called `*Componentname*.ts` . Everything needed for the actual function of the frontend (e.g., search logic (?)) remains in the `*Componentname*.svelte`. That will hopefully make it easier to replace it as the frontend matures. 
- still missing: 
  - separate page/component to display children data for each child --> separate PR
  - some remaining styling issues in `CardDisplay` and `GalleryDisplay`
  - milestone itself is only integrated via a separate page. It's a rather complicated component with a lot of internal functionality that made imho too cumbersome to integrate and adjust it to fit into the new scheme. 

- Sonarcloud is complaining because of the milestone dummy data being copied around.


lkeegan (approved):
lgtm!

PR #107, opened 2024-10-15
Refactor localisation

### LINKED ISSUE #79: Consolidate localisation
Currently we have two more-or-less independent localisation methods:

- frontend
  - uses svelte-i18n library
  - locales are currently hard-coded
  - json files with translations are currently also hard-coded
  - cannot be modified from admin interface, need to modify frontend source code
- backend
  - languages and translations are all stored in the database
  - can/will be modifiable from admin interface

Ideally would should combine them and allow all translations to be edited from the admin interface:

- backend
  - add a table with i18n translations
  - e.g.
    - lang_id
    - key
    - value
  - provide endpoint to return this a json
- frontend
  - should fetch list of locales and json with all translations from the backend via API calls
    - see https://github.com/kaisermann/svelte-i18n/blob/main/docs/Getting%20Started.md#32-asynchronous
  - admin interface should allow editing of the translations

lkeegan:
- User frontend
  - Default language is german with text stored in locales/de.json
    - This is part of the frontend build and so cannot be edited from the backend
  - All other languages are no longer part of the git repo
    - These are fetched as json files from the server
    - These can be edited from the backend
- Admin frontend
  - Add interface for editing translations of frontend text
- Backend
  - Add endpoint to update an 18n translation json file
- resolves #79


MaHaWo (approved):
lgtm

PR #114, opened 2024-10-15
Add Children to backend

### LINKED ISSUE #111: add Children to backend
We need children logic, data and endpoints in the backend:

Child
- id
- user_id
- name
- photo? <-- note: shouldn't be a public image file!
- date of birth

CRUD endpoints authenticated by user with access of children of this user

lkeegan:
- add Child model and routes
  - only the user who owns the child can access these routes
- add STATIC_FILES_PATH to settings
  - this is where non-public files are stored, e.g. child images
- fix user dependency injection in tests
- add public client with no user authentification to tests
- resolves #111


MaHaWo (approved):
lgtm

Issue #109, opened 2024-10-15
add email to backend


lkeegan:
Currently in our UserManager class the methods for forgot password and email confirmation after signup just print a message to the logs with the generated token:

https://github.com/ssciwr/mondey/blob/d4ea8f85283c81e378988904473b4a3d4cd2e9b7/mondey_backend/src/mondey_backend/users.py#L23-L38

We need to modify them to send an email to the user.

I've previously set up a send-only email server for websites using this docker image: https://hub.docker.com/r/boky/postfix

But maybe there is already an email server / account associated with mondey.de that we can/should use instead?

lkeegan:
For now we should just use a send-only mail server

PR #115, opened 2024-10-15
Add backend connection to user login

MaHaWo:
- make use of backend functionality in login, profile popover and user registration 
- add localization where needed
- add 'success' page to user registration
- add error handling where necessary
- make better usage of forms 
- wrt roles:  Currently a user can request a role upon registration, and then would get an e-mail with further instructions on how to get that role (provide verification of a research institution or special admin access token or whatever). The rest of the verification process would then be: 
    - normal observer: click link and be done, as normal 
    - researcher: verify with institution? 
    - admin: verify with admin token (whatever that may be) obtained from an existing admin?
    - currently, only the role request is implemented in the frontend. 
    - is there a standard process for this? 
- Question: does the role have to go into the login too? or should this be automatically determined from the database? ( I assume the latter)
- missing currently: 
    - english translations 
    - users remain unverified #109 
    - no 'reset password' functionality yet in itself. that would have to happen in the user landing page under an additional tab? (can be a late addition imho...)
    - no integration with the admin user stuff yet.


frontend/src/lib/stores/userStore.ts
```diff
@@ -86,4 +88,6 @@ async function hash(input: string): string {
 	return hash;
 }
 
-export { createDummyUser, hash, users, UserStore, type UserData, type UserList };
+const currentUser = writable(null as unknown as UserRead);
```
lkeegan, line 91: why `unknown`? isn't it either null or UserRead? i.e.g `null as null | UserRead`?

frontend/src/lib/components/UserRegistration.svelte
```diff
@@ -1,165 +1,157 @@
...
+			props: {
+				label: $_('registration.role'),
+				type: 'text',
+				placeholder: $_('registration.selectPlaceholder'),
+				required: true,
+				id: 'role',
+				items: [
+					$_('registration.observerRole'),
+					$_('registration.researcherRole'),
+					$_('registration.adminRole')
```
lkeegan, line 97: We can discuss with Sabina, but my 2cents:

- registration should just be username/password and assume normal users, which is presumably the vast majority of users
- researchers who want access should just sign up, then send an email requesting access and have an admin approve them or not
- the admin interface should have a list of all users with the option to make them researchers or admins

frontend/src/lib/components/UserLogin.svelte
```diff
@@ -1,102 +1,101 @@
...
+		const returned = await usersCurrentUser();
+		if (returned.error) {
+			console.log('Error getting current user: ', returned.error.detail);
+			currentUser.set(null as unknown as UserRead);
+			return returned.error.detail;
+		} else {
+			console.log('Successfully retrieved active user');
+			// TODO: remove this in the final version, this sets verified to true
+			// to emulate a successful user registration
+			returned.data.is_verified = true;
```
lkeegan, line 26: I think it makes more sense to make the backend set is_verified=true on signup for now until #109 is implemented

lkeegan (approved):
lgtm - some minor suggestions

frontend/src/lib/components/UserLogin.svelte
```diff
@@ -1,102 +1,101 @@
...
-
-		const user = await users.fetchWithCredentials(
-			data[0].value,
-			await hash(data[1].value),
-			data[2].value
-		);
+		const authReturn = await authCookieLogin(loginData);
 
-		console.log(user);
+		// forget user data again to not have a plain text password lying around in memory somewhere
```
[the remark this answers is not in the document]
MaHaWo: idk if it's a known issue, but I felt uncomfortable having a plain text password lying around in memory, it's really just that.  They will have to type it again, but they would have to do that anyway? Unless it has a11y implications, I wouldn´t judge it too much of a hassle to type the mail again if it's wrong.

frontend/src/lib/stores/userStore.ts
MaHaWo: hm, I was following error tips of the editor at this point, because conversion to UserRead gets flagged as an error and they get overwhelming at some point... I agree that it's unintuitive...

frontend/src/lib/components/UserRegistration.svelte
MaHaWo: would wait for Sabina's input then and change it accordingly then.

frontend/src/lib/components/UserLogin.svelte
MaHaWo: ok... did that now in `on_after_register` in the users.py file. find entry in database corresponding to the passed `user` object, set the verified flag -> write back to database.

PR #118, opened 2024-10-18
Add userdata backend functionality

MaHaWo:
- add models for answers to user questions 
- adds endpoints for put and get requests. No full CRUD because not required afais  
- adds tests 
- adjusts frontend respectively. Because the backend user questions part is not fully functional atm and a fix would have increased complexity and extend of this PR significantly, this currently still uses the hardcoded questions from the frontend page and translates them into the correct structure. This incurs a certain amount of temporary logic to do so, which will go once the admin backend is fixed (next task). 
- adds translations where necessary. user questions are not translated b/c the aforementioned problem. 
- add `required` and `disabled` fields where necessary 
- there's still a lot of debug output in there that will go with the next PR probably 
- stupid hack for making the page rerender: `data = [...data] because of svelte<4 not tracking nested references for rerendering. svelte 5 runes to the rescue?
- the SQLModel models can perhaps be improved, I got a little confused about when and how the database index is added and wanted to get this to a functional state asap with the idea that it might change again, so the primary key and the `question_id` are currently separate, even though the could be the same functionally.


mondey_backend/src/mondey_backend/routers/users.py
```diff
@@ -152,4 +156,59 @@ def update_milestone_answer(
...
+
+        user_answers = [
+            UserAnswerPublic(
+                answer=answer.answer,
+                non_standard=answer.non_standard,
+                question_id=answer.question_id,
+            )
+            for answer in answers
+        ]
+        return user_answers
```
lkeegan, line 176: this shouldn't be required - you can just return answers and fastapi will return a list[UserAnswerPublic] as you specified in the reponse_model

mondey_backend/src/mondey_backend/routers/users.py
```diff
@@ -152,4 +156,59 @@ def update_milestone_answer(
...
+
+    @router.put("/user-answers/", response_model=list[UserAnswerPublic])
+    def update_current_user_answers(
+        session: SessionDep,
+        current_active_user: CurrentActiveUserDep,
+        new_answers: list[UserAnswerPublic],
+    ):
+        if session is None:
+            raise HTTPException(401, detail="no session")
```
lkeegan, line 185: is this needed - i.e. can session be none at this point?

mondey_backend/src/mondey_backend/models/questions.py
```diff
@@ -52,3 +52,40 @@ class UserQuestionAdmin(SQLModel):
...
+class UserAnswerPublic(SQLModel):
+    """
+    External data model for UserAnswers
+
+    Parameters
+    ----------
+    SQLModel : Pydantic model  basic sqlmodel pydantic type
+    """
+
+    answer: str
```
lkeegan, line 89: we could consider adding `id` here - it would make the update logic it more robust (see below)

mondey_backend/src/mondey_backend/routers/users.py
```diff
@@ -152,4 +156,59 @@ def update_milestone_answer(
...
+        else:
+            for new_answer in new_answers:
+                session.add(
+                    UserAnswer(
+                        user_id=current_active_user.id,
+                        question_id=new_answer.question_id,
+                        answer=new_answer.answer,
+                        non_standard=new_answer.non_standard,
+                    )
+                )
```
lkeegan, line 209: this logic looks a bit fragile - e.g. what happens if a new question is added by the admins?
Assuming we add the id to UserAnswerPublic then we could just iterate the provided answers and for each either get and update the existing one in the database, or if it doesn't exist create one, something like (untested):
```python
    for new_answer in new_answers:
        db_new_answer = session.get(UserAnswer, new_answer.id)
        if not db_new_answer:
            db_new_answer = UserAnswer.model_validate(new_answer, update={"user_id": current_active_user.id})
            session.add(db_new_answer)
        else:
            for key, value in new_answer.model_dump().items():
                setattr(db_new_answer, key, value)
```

frontend/src/lib/components/UserDataInput.svelte
```diff
@@ -1,117 +1,166 @@
...
+				element.props.disabled = true;
+			}
+			dataIsCurrent = true;
+			data = [...data]; // TODO: forces a rerender. need a better way to do this
 		}
 	}
 
-	let userData: UserData;
-	let userID: string;
+	// README: is storing userID a problem?
```
lkeegan, line 90: why do we need to store it?

lkeegan (commented):
looks good - made some suggestions for the backend part. I didn't look too much into the frontend as you mentioned it would get refactored.

mondey_backend/src/mondey_backend/routers/users.py
MaHaWo: ok, changed that.

mondey_backend/src/mondey_backend/routers/users.py
MaHaWo: changed that into a tuple key as we discussed yesterday.

mondey_backend/src/mondey_backend/routers/users.py
MaHaWo: hm, it did make things a bit easier, but since the usage of userID has reduced by now to the `submitData` function I retrieve it there now.

mondey_backend/src/mondey_backend/models/questions.py
MaHaWo: changed accordingly as discussed yesterday

mondey_backend/src/mondey_backend/routers/users.py
MaHaWo: what guarantees(!) us that it can not? removing this afais would result in a  code 500 error without useful message.

frontend/src/lib/components/UserDataInput.svelte
MaHaWo: I managed to do things without it, this was a leftover from an earlier iteration...

PR #127, opened 2024-10-24
Add de/en languages to the database on startup if there are no languages

### LINKED ISSUE #126: Should create `de` language on startup if languages table is empty in the database
..we assume it exists in the frontend anyway

lkeegan:
- ensures the languages database always contains the rows
  - 1 de
  - 2 en
- don't allow admins to delete either of these languages
  - no use case where this would be required
  - and could make a mess in production if it happened by accident
- remove delete button from admin interface for these languages
- resolves #126

### DIFF
--- a/frontend/src/lib/components/Admin/Languages.svelte
+++ b/frontend/src/lib/components/Admin/Languages.svelte
@@ -73,12 +73,12 @@
 					</TableBodyCell>
 					<TableBodyCell>
 						{#if lang_id > 2}
-						<DeleteButton
-							onclick={() => {
-								currentLanguageId = `${lang_id}`;
-								showDeleteModal = true;
-							}}
-						/>
+							<DeleteButton
+								onclick={() => {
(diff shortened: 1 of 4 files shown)


MaHaWo (commented):
changes look good, but two things came up while using the frontend: 
- for me, `en` does not show in the frontend: only `de` is shown. when I add another language, that behaves like the second language and cannot be deleted anymore 
- there seems to be an issue with the uniqueness of database entries: adding the same language multiple times does not produce separate entries in the frontend page, however, it seems to create multiple database entries, because I have to click on delete as often as I added it to make it disappear again. Can the database table for languages have only a primary key that would be the `lang_id` string perhaps? that should solve the problem I think (similar issue to the one we had with the user answers it seems).

MaHaWo:
discussion IRL:  
- first is an issue that arises when starting from a non-empty database 
- second issue will become new pr

PR #129, opened 2024-10-24
Use 2-char lang code as Language primary key

### LINKED ISSUE #128: Make Language primary key the 2-char code itself
- enforces uniqueness of language codes
- involves some refactoring elsewhere in backend & fronted

lkeegan:
- enforces uniqueness of languages
- get languages now returns a list of these lang codes
- remove langStore from frontend, instead we can now directly use the i18n-svelte library:
  - `$locale` for current lang_id
  - `$locales` for list of available lang_ids
- update frontend and admin frontend accordingly
- note: involves many changes to the database, probably easiest to just delete milestones.db
- resolves #128


MaHaWo (approved):
lgtm

PR #137, opened 2024-10-29
Replace prettier/eslint with biome

### LINKED ISSUE #124: Replace eslint/prettier with biome
Prettier doesn't integrate well with pre- commit, so we run it in ci.
This is much less convenient than having it in precommit.

biome is a fast replacement for Eslint and prettier and comes with a pre-commit hook, looks

lkeegan:
- add biome to pre-commit config
- add biome config: for now exclude all failing linter rules
  - in the future we can remove these one by one and fix the code
- remove pnpm format from codegen ci
- remove pnpm format commands
- remove prettier/eslint configs and installs
- resolves #124


MaHaWo (approved):
:+1:

PR #138, opened 2024-10-29
Add childquestions backend integration and refactor datainput datastructure

### LINKED ISSUE #112: add ChildQuestions
Likely exactly the same as UserQuestion - probably should make a Question then have them both inherit from it.

### LINKED ISSUE #113: Add ChildQuestionAnswer and UserQuestionAnswer
We need to add these to store the answers to questions + CRUD endpoints

MaHaWo:
## What it does
- integrates backend into `ChildrenDataInput`  and `ChildrenGallery` frontend component 
- add `delete_child_image` endpoint to user backend.
- makes more use of svelte-5 runes in components related to child and data input.
- various smaller fixes mostly related to types
- modifies the `FileInput` component to work better with the backend system
- functionally unifies the question editing dialog for user and child questions -> `EditQuestions.svelte` , `Question.svelte`
- integrate question edition into admin page.
- add creation of new child with necessary data to the children datainput dialog. Submit and markup section handle the data that is related to the child and the data that belongs to questions separately as discussed. 
- make the answer response for children a `dict[id, answer]` instead of `list[answer]`. This is a little easier to maintain in the frontend. Arguably should be the same for user answers, but not done here to not conflate things too much.
- delete frontend code that emulated backend
- delete childstore and replace with one that just stores the current child's id or null if there is none. Still using stores instead of runes here, imho they are easier to use with less code for what I need them for. Might be necessary to migrate at some point though. 
- adjust python tests where necessary. 
- the diff view won´t be much help for `ChildrenGallery.svelte` and `ChildrenRegistration.svelte` as especially for the latter, most of the code changed.
## Questions
- do we need the 'order' parameter in the questions or can this role be fulfilled by the id? 
- child image handling is a little suboptimal imho: 
  - Image deletion is a bit complicated: Other question elements can be deleted by providing an empty value as input. For image files, this is not  as straight forward since it's stored separately. I don´t want to do anything implicit à la "if empty string then and child had a file previously, then delete". So I added an explicit delete button if there is an image already. This is displayed in addition to the upload dialog. To make all this work and activate and deactivate at the right time, 2 new boolean flags are needed. 
  - this might all simplify a bit if it turns out during milestone integration that we should store the whole child object as `$currentChild` and not just it's id. I was hesitant to do this right now though, because it stores a lot of data in the frontend.  

closes #113
closes #112


MaHaWo:
found a bug that slipped through the net. need to fix that before review. sry

MaHaWo:
ok, I think I'm done @lkeegan. feel free to start whenever you have time.

frontend/src/lib/components/ChildrenGallery.svelte
```diff
@@ -1,89 +1,159 @@
...
+				image: null,
+			},
+			...childrenData,
+		];
 	}
+	return data;
 }
 
-function searchRemarks(data: any[], key: string): any[] {
+function createStyle(data: any[]) {
```
lkeegan, line 75: How about if GalleryDisplay took a list of items, where each item has
```js
{data = ..., component = ..., componentProps = ...}
```
rather than taking a separate list for each of these?
It seems (at least to me) that would reduce the need for this extra logic / data manipulation?

mondey_backend/src/mondey_backend/routers/users.py
```diff
@@ -95,12 +95,14 @@ async def get_child_image(
...
-        child = child = get_db_child(session, current_active_user, child_id)
+        child = get_db_child(session, current_active_user, child_id)
         image_path = pathlib.Path(
             f"{app_settings.PRIVATE_FILES_PATH}/children/{child.id}.jpg"
         )
         if not image_path.exists():
-            raise HTTPException(404)
+            raise HTTPException(404, detail="Image not found")
+        if child.has_image is False:
+            raise HTTPException(404, detail="Child does not have an image")
```
lkeegan, line 105: rather than raising an error, maybe we should have a default blank avatar image that we return if there isn't an image for the child?

mondey_backend/src/mondey_backend/routers/users.py
```diff
@@ -117,6 +119,22 @@ async def upload_child_image(
...
+        child = get_db_child(session, current_active_user, child_id)
+        image_path = pathlib.Path(
+            f"{app_settings.PRIVATE_FILES_PATH}/children/{child.id}.jpg"
+        )
+        if not image_path.exists():
+            raise HTTPException(404, detail="Image not found")
+        else:
+            child.has_image = False
+            session.commit()
+            image_path.unlink()
```
lkeegan, line 135: I'd suggest not checking if the file exists, and just doing the else part of this, with .unlink(missing_ok=True) - if the file is already deleted for whatever reason then that's actually fine, and we shouldn't abort setting child.has_image to false imo

lkeegan (approved):
lgtm - some minor comments/suggestions.

re questions:
1. we need `order` if we want the admins to be able to change the order of the questions, if they are fine with them coming in the order that they were added to the database then yes we can remove it (same goes for milestonegroup)
2. image handling: do we need to be able to delete the image - or is just allowing the user to upload a different one sufficient?
3. yes I would store the child object

frontend/src/lib/components/ChildrenGallery.svelte
MaHaWo: yes, this is my intention, but I didn´t want to do this here as it will result in refactoring in at least 3 (frontpage, childrengallery, milestones) different places  (everywhere gallery is used)

MaHaWo: issue: #151

mondey_backend/src/mondey_backend/routers/users.py
MaHaWo: good point.

MaHaWo:
thanks for the input. Regarding image deletion: 
- I think we should be able to delete it: 
  - users might just change their mind. 
  - user might upload an image but later find out they were not allowed to/ did so on accident or are not allowed anymore.
  - it might be more convenient to have no images when the number of observed children grows - kindergarden or the like.
  => therefore, option to delete image would be nice to have.

mondey_backend/src/mondey_backend/routers/users.py
MaHaWo: I would prefer to do this later, however.

MaHaWo:
I added the default image for image-less children now, too.

PR #139, opened 2024-10-29
Design of the main page

Majapur1:
- Cosmetic changes of navigation bar (_removing the HOME button, centering the buttons and increasing the space from the top edge of the page_);

- Changing a primary color.

### DIFF
--- a/frontend/src/routes/+layout.svelte
+++ b/frontend/src/routes/+layout.svelte
@@ -19,12 +19,11 @@
 
 <Navbar>
 	<NavBrand href={base}>
-		<img src={logo_light} class="block h-16 dark:hidden" alt="MONDEY Logo" />
-		<img src={logo_dark} class="hidden h-16 dark:block" alt="MONDEY Logo" />
+		<img src={logo_light} class="block mt-6 h-12 dark:hidden" alt="MONDEY Logo" />
+		<img src={logo_dark} class="hidden mt-6 h-12 dark:block" alt="MONDEY Logo" />
 	</NavBrand>
 	<NavHamburger />
-	<NavUl ulClass="flex space-x-4 text-lg ">
-		<NavLi href={base} active={true}>Home</NavLi>
+	<NavUl ulClass="flex items-center mt-8 space-x-14 text-lg ">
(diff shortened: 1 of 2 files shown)


frontend/src/routes/+layout.svelte
```diff
@@ -19,12 +19,11 @@
...
 	<NavBrand href={base}>
-		<img src={logo_light} class="block h-16 dark:hidden" alt="MONDEY Logo" />
-		<img src={logo_dark} class="hidden h-16 dark:block" alt="MONDEY Logo" />
+		<img src={logo_light} class="mt-6 block h-12 dark:hidden" alt="MONDEY Logo" />
+		<img src={logo_dark} class="mt-6 hidden h-12 dark:block" alt="MONDEY Logo" />
 	</NavBrand>
 	<NavHamburger />
-	<NavUl ulClass="flex space-x-4 text-lg ">
-		<NavLi href={base} active={true}>Home</NavLi>
+	<NavUl ulClass="flex items-center mt-8 space-x-14 text-lg ">
```
lkeegan, line 26: Have you checked what this looks like on a mobile device screen size?
It might make sense to only have these additional margins and x spacing on large screens - and this is easy to do as tailwind is [mobile first](https://tailwindcss.com/docs/responsive-design), so you can do something like e.g. `mt-2 md:mt-8` to have a smaller margin by default (i.e. on mobile devices), then apply the larger margin for the `md` breakpoint (>768px width)

lkeegan (commented):
One suggestion, otherwise looks good, thanks.

Also a general note: in github "assign" is typically used to say who will work on this issue, when you want someone to review and merge your PR you can instead ask for a "review"

MaHaWo:
Is the bottom part you had in your design still missing or should this not be part of it anymore?

frontend/src/routes/+layout.svelte
Majapur1: Thanks! I've just fixed that. Now it looks better with all screen sizes.

Majapur1:
> Is the bottom part you had in your design still missing or should this not be part of it anymore?

Still missing. Should I make a full page and only then send a pull request?

PR #140, opened 2024-11-04
Add expected age to Milestone

### LINKED ISSUE #124: Replace eslint/prettier with biome

lkeegan:
- Milestone now has `expected_age_months`: int with default of 12
- Admin interface allows admin to set this value using a slider
- resolves #124

Also restored `admin.title` in de locale as it was already used for 'Titel', added `admin.heading` for 'Administration'


MaHaWo (commented):
some of the comments in #142 apply here as well I think, so won´t repeat them here. approved.

PR #142, opened 2024-11-04
Return age-appriopriate Milestones

### LINKED ISSUE #132: Get milestones should return age-appropriate milestones
For a child with age `x months`, get milestones should return all milestones for which

`x - 6 < `Milestone.expected_age_months` < x + 6`

lkeegan:
- `/milestone-groups` endpoint now requires a `child_id`
- it returns only the milestones with expected age within +/-6 months of the child age for each group
- remove previous endpoint that returned a single milestone group by id as it is not needed
- refactor children test fixture to reduce duplication
- add some missing type annotations
- update demo milestone route in frontend accordingly
- resolves #132

depends on #140


frontend/src/lib/components/Admin/EditMilestoneModal.svelte
```diff
@@ -75,6 +76,10 @@ export async function saveChanges() {
 				{/each}
 			</div>
 		{/each}
+		<div class="mb-5">
+		<Label>{`${$_("admin.expected-age")}: ${milestone.expected_age_months}m`}</Label>
+		<Range id="expected-age-months" min="1" max="72" bind:value={milestone.expected_age_months} />
```
MaHaWo, line 81: when I add a milestone and then try to edit it, this doesn´t show up because the `milestone` variable is still `null`. Didn´t track down what goes wrong there in detail though...

frontend/src/routes/milestone/+page.svelte
```diff
@@ -20,7 +20,12 @@ let milestoneAnswerSession = $state(
 let childId = $state(undefined as number | undefined);
```
MaHaWo, line 21: This route only displays an empty page for me?

frontend/src/lib/components/AdminPage.svelte
```diff
@@ -21,7 +21,6 @@ onMount(async () => {
 </script>
 
 <Tabs tabStyle="underline" class="w-full">
```
MaHaWo, line 23: in line 30, there is the `MilestoneGroups` component loaded into the tabitem. I'm a little confused by this distinction now. Is this correct, or rather, is the `Milestone` part missing here?

mondey_backend/src/mondey_backend/routers/milestones.py
```diff
@@ -35,17 +40,35 @@ def get_milestones(
...
+        child_age_months = get_child_age_in_months(child)
         milestone_groups = session.exec(
-            select(MilestoneGroup).order_by(col(MilestoneGroup.order))
+            select(MilestoneGroup)
+            .order_by(col(MilestoneGroup.order))
+            .options(
+                lazyload(
+                    MilestoneGroup.milestones.and_(
+                        (
+                            child_age_months
```
MaHaWo, line 61: is the `<=` / `>=` on both ends intentional here?

MaHaWo (approved):
A few questions , otherwise looks good.

PR #144, opened 2024-11-06
Refactor userland pages 

MaHaWo:
- removes much of the old datastructures
- makes more use of runes
- adds a small `DateInput` data input component
- updates `fileupload`  component 
- rebuilds userregistration with #await block


MaHaWo:
included comments. merging asap

PR #145, opened 2024-11-06
Changing the card gallery to one card on the main page

Majapur1:
- Added one card to the home screen instead of the gallery (it stayed within the gallery, it is possible to add more cards if needed).

### DIFF
--- a/frontend/src/lib/components/Frontpage.svelte
+++ b/frontend/src/lib/components/Frontpage.svelte
@@ -3,8 +3,8 @@
 import { base } from "$app/paths";
 
 import CardDisplay from "$lib/components//DataDisplay/CardDisplay.svelte";
-import Card from "$lib/components/FrontPageCard.svelte";
 import GalleryDisplay from "$lib/components/DataDisplay/GalleryDisplay.svelte";
+import Card from "$lib/components/FrontPageCard.svelte";
 
 export let getStarted = "";
 
--- a/frontend/src/lib/components/Frontpage.svelte
+++ b/frontend/src/lib/components/Frontpage.svelte
@@ -3,7 +3,7 @@
 import { base } from "$app/paths";
 
 import CardDisplay from "$lib/components//DataDisplay/CardDisplay.svelte";
-import Card from "$lib/components/Card.svelte";
+import Card from "$lib/components/FrontPageCard.svelte";
 import GalleryDisplay from "$lib/components/DataDisplay/GalleryDisplay.svelte";
 
 export let getStarted = "";
@@ -13,7 +13,7 @@ export let items = [];
 const props = {};
 </script>
 
-<Card
+<FrontPageCard
 />
 
 <GalleryDisplay
--- a/frontend/src/lib/components/Frontpage.svelte
+++ b/frontend/src/lib/components/Frontpage.svelte
@@ -4,7 +4,7 @@ import { base } from "$app/paths";
 
 import CardDisplay from "$lib/components//DataDisplay/CardDisplay.svelte";
 import GalleryDisplay from "$lib/components/DataDisplay/GalleryDisplay.svelte";
-import Card from "$lib/components/FrontPageCard.svelte";
+import FrontPageCard from "$lib/components/FrontPageCard.svelte";
 
 export let getStarted = "";
 
--- a/frontend/src/lib/components/Frontpage.svelte
+++ b/frontend/src/lib/components/Frontpage.svelte
@@ -3,71 +3,21 @@
 import { base } from "$app/paths";
 
 import CardDisplay from "$lib/components//DataDisplay/CardDisplay.svelte";
+import Card from "$lib/components/Card.svelte";
 import GalleryDisplay from "$lib/components/DataDisplay/GalleryDisplay.svelte";
 
 export let getStarted = "";
 
-export let items = [
-	{
-		header: "Was ist Mondey?",
-		summary:
-			"Mondey ist ein wissentschaftlich geprüftes Programm zure Dokumentation der Entwicklung von Kindern bis 6 Jahren.",
-		href: `${base}/info`,
-		button: "Mehr Info",
-	},
-	{
-		header: "Wozu ist Mondey gut?",
-		summary:
-			"Anhand ihrer Bewertungen der Fähigkeiten des Kindes erhalten sie Feedback zum Entwicklungsstand des Kindes und können so frühzeitig Fördermaßnahmen einleiten. Dies folgt einem übersichtlichen Ampelsystem.",
-		href: `${base}/info`,
-		button: "Mehr Info",
-	},
-	{
-		header: "Was umfasst Mondey?",
-		summary:
-			"Mondey umfasst unterschiedliche Entwicklungsbereiche wie von Kindern im Alter von 0 bis 6 Jahren. Dazu gehören unter anderem Grob-und feinmotorik, Wahrnehmung, Denkne, Sprache und Soziale Beziehungen.",
-		href: `${base}/info`,
-		button: "Mehr Info",
-	},
-	{
-		header: "Wie funktioniert Mondey?",
-		summary:
-			"Sie bewerten wie gut das Kind bestimmte Alltagshandlungen durchführen kann mit Hilfe einer Liste von Fragen.",
-		href: `${base}/info`,
-		button: "Mehr Info",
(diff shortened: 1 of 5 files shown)


frontend/src/lib/assets/baby-flower.png
lkeegan, line 1: this png file is huge! try converting it to a jpg :-) also I don't see it being used anywhere?

frontend/src/lib/components/DataDisplay/CardDisplay.svelte
```diff
@@ -27,36 +27,34 @@ export let styleProps = {
 	imgClass="max-md:hidden object-scale-down"
 	href={data.button ? null : data.href}
 	class={data.button
-		? 'm-2 max-w-prose items-center  text-gray-700 dark:text-white'
+		? 'm-2 max-w-prose min-w-full items-center bg-[#a4d2d0] text-gray-700 dark:text-white'
```
lkeegan, line 30: I think the CardDisplay and GalleryDisplay components are used in multiple places, not just on the front page (@MaHaWo?).
It might make more sense for you to directly use a single `<Card>` element on the frontpage which you can customize however you like, rather than modifying these components?

lkeegan (changes requested):
thanks - made a couple of suggestions

frontend/src/lib/assets/baby-flower.png
MaHaWo: second that. Do we need it at all?

frontend/src/lib/components/DataDisplay/CardDisplay.svelte
MaHaWo: yes, they are indeed, I would also suggest to use `Card` directly if you need only one of them. also, I would suggest using colors that are defined centrally in `tailwind-config.ts`, because a unified style will depend on the color hexcode being copied everywhere we need it, which is brittle

frontend/src/lib/components/DataDisplay/GalleryDisplay.svelte
```diff
@@ -108,7 +108,7 @@ $: filteredComponentProps = filteredItems.map((item) => {
 	{/if}
 
 	<Gallery
-		class="grid w-full grid-cols-1 justify-center gap-8 p-4 sm:grid-cols-1 md:grid-cols-2 lg:grid-cols-3 xl:grid-cols-4"
+		class="w-full justify-items-center"
```
MaHaWo, line 111: same here. While the styling isn´t final, some of the other components that use the `GalleryDisplay` depend on it being styled as originally defined and their styling will break if it is changed.

frontend/src/lib/components/Frontpage.svelte
```diff
@@ -13,60 +13,10 @@ export let getStarted = "";
 
 export let items = [
 	{
-		header: "Was ist Mondey?",
-		summary:
-			"Mondey ist ein wissentschaftlich geprüftes Programm zure Dokumentation der Entwicklung von Kindern bis 6 Jahren.",
+		header: "Möchten Sie die Entwicklung von Kindern begleiten und fördern?",
```
MaHaWo, line 16: you can use the `svelte-i18n` package to make use of localization. There is currently one default localization file under `frontend/src/locales` which is called `de.json`. You can define a section "frontpage" there and add all the items you need: 

```json 
"frontpage": {
    "heading": "Möchten Sie die Entwicklung von Kindern begleiten und fördern?",
     "summary": "Hier sind Sie genau richtig!",
     "buttonLabel": "Anmeldung"
} 
```

which then can be used in the  `*.svelte` files like so: 

```javascript
import {_} from "svelte-i18n";

export let items = [
       {header: $_("frontpage.heading"), ... }
```
and so on. If you want to use these in the markup section of the component, you have to wrap it in `{}`.

MaHaWo (commented):
a few remarks herer and there.

frontend/src/lib/assets/baby-flower.png
Majapur1: Ok! I've added it for use in the next element. I'll compress it and add it again later

frontend/src/lib/components/DataDisplay/CardDisplay.svelte
Majapur1: Thanks for your help and suggestions! I'll use the Card element in those cases and add other colors to tailwind-config.ts

frontend/src/lib/components/DataDisplay/GalleryDisplay.svelte
Majapur1: Ok, get it.

Majapur1:
- add a card to the home page as a separate element

Majapur1:
> thanks @Majapur1 - but this PR has got a bit confusing to review! (now 57 files changed & merge conflicts with main).
> 
> Am I right in thinking that the only changes you actually want to make here are the changes from your most recent commit?: [d537dbb](https://github.com/ssciwr/mondey/commit/d537dbbee11de89393a9bd41abc095c0f8539fff)

Sorry, I've did it wrong! Yes, you're right, all real changes are there.

d537dbbee11de89393a9bd41abc095c0f8539fff

lkeegan:
@Majapur1 the easiest way for you to update your local branch now to match this one might be to delete it, then check it out again. Alternatively something like `git reset HEAD~80 --hard && git pull` should also work.

frontend/src/lib/components/Frontpage.svelte
```diff
@@ -3,72 +3,22 @@
...
-];
+export let items = [];
 
 const props = {};
 </script>
 
+<Card
+	withSearch={false}
+	itemComponent={Card}
+	componentProps={props}
```
lkeegan, line 19: these are [props](https://svelte.dev/tutorial/svelte/declaring-props) for the GalleryDisplay component - your Card component doesn't have any props, so you don't need to pass anything here
```suggestion
```

frontend/src/lib/components/Frontpage.svelte
```diff
@@ -3,72 +3,22 @@
...
-		href: `${base}/info`,
-		button: "Mehr Info",
-	},
-];
+export let items = [];
 
 const props = {};
 </script>
 
+<Card
```
lkeegan, line 16: Since your Card component has hard-coded contents, `Card` might be a bit too generic of a name, maybe you could rename it to FrontPageCard.svelte or something similarly specific?

lkeegan (commented):
a couple of suggestions, otherwise looks good!

frontend/src/lib/components/Frontpage.svelte
Majapur1: Okey, I was not sure about it. Thanks!

Do I need to delete these lines manually or there's some quick way to fix lines of code with Github and without VSC?

frontend/src/lib/components/Frontpage.svelte
Majapur1: Ok, I get it. Thank you!

Here I have a same question — can I rewrite this line quickly? @lkeegan

frontend/src/lib/components/Frontpage.svelte
lkeegan: Typically you would delete those lines on your local branch, make a commit with an appropriate message, then push.
You can also edit the file using the github website if you like.

frontend/src/lib/components/Frontpage.svelte
lkeegan: You need to rename the Card.svelte file as well - I'd recommend doing this locally, making a commit & pushing. (It sounds like a lot but with some practice it can also be very quick :-))
It might also be possible to rename files from the github website these days, but I've never tried.

Majapur1: Good, that'll help me get through this faster :)

@lkeegan and the last question! How often can I create new branches to accomplish tasks? Is it a problem for if I do it too often or for very small changes?

lkeegan: You should create a new branch for every new feature, it's fine if it is only a small change, in fact small changes are good - it makes it much easier to review the Pull Request if it is small and self-contained.

Majapur1: okey! Thank you for help and guidance!

lkeegan (approved):
lgtm!

PR #158, opened 2024-11-12
Integrate milestones userland with backend

MaHaWo:
- integrate milestone and milestone overview into user landingpage
- remove old dummy code for milestones
- adjust `Milestone` component to the userland scheme (make use of setup function and #await block etc)
- include the categorization according to milestone groups again. Displaying all milestones could potentially become a lot of content, given that there are ~200 milestones that belong to < 10 age categories. This can be changed according to experience or desire by removing the `MilestoneGroup` component from the navigation breadcrumbs and adjusting the `contentStore`. I would ask for feedback from client in this regard.
- breadcrumbs and tabs can now have symbols that are used when the screen becomes too small for written labels to be sensibly displayed.
- add progress status to `MilestoneGroupPublic` which is computed from the currently active answer session. Since this is only an indicator for the user, it is computed on the fly and there is no additional field in the database model for it
- make the layout of the milestone such that it auto adjusts to viewport width
- store data of current child 
- store current milestonegroup 
-  refactoring: 
   - make more use of localization and svelte 5 runes mode
   - adjust breadcrumbs 
   - add navigation button for milestonegroups to child data overview instead of relying on the breadcrumbs for it


frontend/src/lib/components/MilestoneOverview.svelte
```diff
@@ -1,40 +1,39 @@
...
 		return data;
 	} else {
 		return data.filter((item) => {
 			// button label contains info about completion status => use for search
-			if (key === "fertig") {
-				return item.complete === true;
-			} else if (key === "unfertig") {
-				return item.complete === false;
+			if (key === $_("milestone.complete")) {
+			} else if (key === $_("milestone.incomplete")) {
```
lkeegan, line 30: is there some logic missing here?

frontend/src/lib/components/UserLandingPage.svelte
```diff
@@ -18,23 +18,33 @@ import UserDataInput from "./UserDataInput.svelte";
...
-				<TabItem>
+				<TabItem onclick = {() =>{
+					activeTabChildren.set("childrenGallery");
+				}}>
 					<div slot="title" class="flex items-center gap-2 text-lg">
 						<GridPlusSolid size="lg" />
-						Kinder
+						{#if smallScreen === false}
+							{$_("childData.overviewLabel")}
+						{/if}
```
lkeegan, line 47: I think you should be able to use tailwind classes for this, e.g. `class="invisible md:visible`: https://tailwindcss.com/docs/visibility#breakpoints-and-media-queries

frontend/src/lib/stores/childrenStore.svelte.ts
```diff
@@ -0,0 +1,47 @@
...
+		},
+		get name() {
+			return childdata?.name;
+		},
+		get month() {
+			return childdata?.birth_month;
+		},
+		get year() {
+			return childdata?.birth_year;
+		},
```
lkeegan, line 18: is there a benefit to adding these getters vs just providing access to the ChildPublic object?

mondey_backend/src/mondey_backend/models/milestones.py
```diff
@@ -60,8 +60,9 @@ class MilestoneGroup(SQLModel, table=True):
 
 class MilestoneGroupPublic(SQLModel):
     id: int
-    text: dict[str, MilestoneGroupTextPublic]
-    milestones: list[MilestonePublic]
+    text: dict[str, MilestoneGroupTextPublic] = {}
+    milestones: list[MilestonePublic] = []
```
lkeegan, line 64:
```suggestion
    text: dict[str, MilestoneGroupTextPublic]
    milestones: list[MilestonePublic]
```

mondey_backend/src/mondey_backend/routers/milestones.py
```diff
@@ -67,6 +70,42 @@ def get_milestone_groups(
...
             )
         ).all()
-        return milestone_groups
+
+        # compute progress from milestone answers. While this is not
+        # the most efficient way, it is the least intrusive and concentrates
+        # progress reporting in the least amount of code - this function and the
+        # MilestonAnswerPublic model.
+        # The actual database does not need to know about progress - it's
+        # a derived value only for being displayed to the user.
```
lkeegan, line 79: I'd suggest putting this logic in the frontend:
- `progress` is only used by the frontend
- the frontend already has the data needed to calculate it (milestones list and current answer session)
- and it should actually be less code to implement there than here, e.g. something like (untested)
```js
milestones.filter((milestone) => {return `${milestone.id}` in answerSession.answers}).length / milestones.length
```
- this reduces load on the server
- and means we don't need to re-calculate it in the server when the user submits an answer

frontend/src/lib/components/Milestone.svelte
```diff
@@ -124,105 +114,154 @@ function selectAnswer(answer: number | undefined) {
...
+	},
+]);
 </script>
 
+{#await promise}
+<p>{$_("userData.loadingMessage")}</p>
+{:then}
 <div
-	class="border-1 flex flex-col border border-gray-200 bg-white shadow md:max-w-7xl md:rounded-lg dark:border-gray-700 dark:bg-gray-800"
+	class="mx-auto flex flex-col p-4 md:rounded-t-lg"
```
lkeegan, line 183: not sure if it is this line or elsewhere but the styling has some issues for me:

image

lkeegan (commented):
thanks - looks good! made some suggestions & noticed a styling issue for the milestone component

frontend/src/lib/components/MilestoneOverview.svelte
MaHaWo: thanks. I indeed overlooked that at some point :grimacing:

frontend/src/lib/stores/childrenStore.svelte.ts
MaHaWo: I personally find it more intuitive to be able to write child.name, child.month etc instead of child.data.month, and b/c I don´t see a reason why the data model should change it doesn´t seem to be a maintenance dept.

mondey_backend/src/mondey_backend/models/milestones.py
MaHaWo: thx. happened during merge...

mondey_backend/src/mondey_backend/routers/milestones.py
MaHaWo: true, we already need the milestone answer session object.

frontend/src/lib/components/UserLandingPage.svelte
MaHaWo: good point, thx.

MaHaWo: thx. went with `hidden md:inline`  to make the invisible elements not occupy space in the layout.

frontend/src/lib/components/Milestone.svelte
MaHaWo: I didn´t manage to reproduce this so far. was the image expecially large or anything?

MaHaWo: ok, so I experimented with the [`Carousel`](https://flowbite-svelte.com/docs/components/carousel#Component_data)  but didn´t get it work, and eventually gave up. I hope  this works now with the modified tailwind classes and I didn´t overlook some cases.

mondey_backend/src/mondey_backend/routers/milestones.py
MaHaWo: was originally hesitant to put more computation into the frontend, b/c it already has quite a bit to do.

MaHaWo:
@lkeegan  you want to have another look?

lkeegan (approved):
looks good! one suggestion.
also the image sizing could still be improved but we can come back to it later
image

MaHaWo:
would come back to this later to not stall this.

PR #159, opened 2024-11-13
Ensure `MilestoneGroup.milestones` is sorted correctly

lkeegan:
- sort milestones list by ascending order of `Milestone.order`
- update test data to include ordering that is not the same as the order of ids


MaHaWo (approved):
:+1:

PR #161, opened 2024-11-13
Add admin ordering of items

### LINKED ISSUE #141: order milestones by expected age?
Currently the milestones have an `order` integer which can be used to determine the order in which they are returned. (If the order is the same for two milestones, which it is currently for all of them, then the order of creation is used)

However, now that we have an expected age for each milestone, maybe it would be an improvement to order them by that, i.e. in ascending order of age/difficulty?

This seems like a natural ordering, and would allow us to remove the `order` from the Milestones.

lkeegan:
- backend
  - add /*/order endpoints to re-order lists of
    - milestones
    - milestone-groups
    - child-questions
    - user-questions
- admin frontend
  - add "reorder" button to all lists of items that can be ordered
  - add OrderItemsModal dialog to re-order items using drag&drop
  - add svelte-dnd-action library for drag&drop functionality
- resolves #141


mondey_backend/src/mondey_backend/models/utils.py
```diff
@@ -1,12 +1,18 @@
...
 
+from pydantic import BaseModel
 from sqlalchemy import String
 from sqlalchemy.orm import attribute_keyed_dict
 from sqlalchemy.orm import relationship
 from sqlmodel import Field
 from sqlmodel import Relationship
 
 
+class ItemOrder(BaseModel):
```
MaHaWo, line 11: can´t this go into models.py?

MaHaWo (approved):
lgtm

PR #168, opened 2024-11-14
Allow admins to view and edit users

### LINKED ISSUE #164: Add Users tab to admin interface
Admins should be able to see and edit a list of users

lkeegan:
- backend
  - add /admin/users/ endpoint to get a list of users
  - refactor admin endpoints into separate files for languages, milestones, users, questions
  - refactor tests to also create a temporary in-memory users database for each test
    - add pytest-asyncio to allow use of async fixtures
  - refactor tests to be fully independent of each other
    - new app & new temp dirs for each test
    - makes test suite slower to run but easier to maintain
  - remove DATA_FILES_PATH as we already have STATIC_FILES_PATH and PRIVATE_FILES_PATH
- admin-frontend
  - add users tab with table of users whose permissions can be edited
- resolves #164


MaHaWo (approved):
maybe it would be better to remove the default image entirely here.

lkeegan:
> maybe it would be better to remove the default image entirely here.

true, but I'll do it in a separate PR, with all the unrelated refactoring this PR has already diverged enough from what its title claims it is about!

PR #170, opened 2024-11-15
add an element "About" to the front page

Majapur1:
- Information about MONDEY-project (text + image)


Majapur1:
@lkeegan What could be the reason why I can't see the image I added?

lkeegan:
> @lkeegan What could be the reason why I can't see the image I added?

In the script part you need to do `import babyFlower from "$lib/assets/babyFlower.jpg";`

Then you can use this babyFlower variable for img, e.g. `<Card img={babyFlower} ...>`

Majapur1:
@lkeegan ok, thanks!

PR #171, opened 2024-11-15
160 fix user login

MaHaWo:
- fix runes mode usage 
- make use of runes wherever appropriate 
- activate biome linter rules 
- fix the uncovered quality issues as far as possible, with 2 exceptions: 
   - Flowbite-svelte components apparently don´t play nice with svelte 5 when being passed as arguments to other components and being used with out `<svelte:component ... />` 
   - `$locale` is flagged as possibly `null | undefined` when using it as index variable, but I don´t see how this can happen in the current system.
- delete `AbstractContent` component that is no longer needed 
- put tests for table functionality in separate file. this might go soon, but I'm hesitant to delete it.


frontend/src/lib/components/UserLandingPage.svelte
```diff
@@ -42,8 +39,7 @@ onMount(async () => {
 						this={componentTable[$activeTabChildren]}
 					/>
 				</TabItem>
-
-				{#if get(currentUser)?.is_superuser}
+				{#if user.data?.is_superuser}
```
lkeegan, line 42: since you already did an {#if user.data} I don't think we need ?. here

frontend/src/lib/components/UserLogin.svelte
```diff
@@ -1,30 +1,27 @@
...
 		console.log("Error getting current user: ", returned.error.detail);
-		currentUser.set(null);
 		return returned.error.detail;
-	} else {
-		console.log("Successfully retrieved active user");
-		currentUser.set(returned.data as UserRead);
-		return "success";
 	}
+	console.log("Successfully retrieved active user");
+	user.data = returned.data as UserRead;
```
lkeegan, line 23: if you also check that user.data is not undefined above, I don't think you need the as UserRead cast here

lkeegan (approved):
nice cleanup job, thanks! some minor suggestions

frontend/src/lib/components/ChildrenRegistration.svelte
```diff
@@ -349,18 +351,18 @@ async function submitData(): Promise<void> {
 							bind:value={answers[element.id].answer}
 							bind:additionalValue={answers[element.id]
 								.additional_answer}
-							label={element?.text[$locale].question}
+							label={element?.text?.[$locale]?.question ?? ''}
```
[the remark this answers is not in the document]
MaHaWo: it is indeed, I removed a couple of them again, ignoring the raised issues because it doesn´t make much sense that `$locale` would be `null` or `undefined`. but the top level wrapper is a nice idea.

MaHaWo:
thanks for the hints. will merge when ready.

PR #172, opened 2024-11-15
Refactor image files handling

### LINKED ISSUE #169: Remove default child image
If there is no child image uploaded, then no image should be displayed

lkeegan:
- backend
  - add sub-folders to static files directory
    - child images: PRIVATE / children / {child_id}.jpg
    - milestone images: STATIC / m / {milestone_image_id}.jpg
    - milestone group images: STATIC / mg / {milestone_group_id}.jpg
    - i18n translations: STATIC / i18n / {language_id}.json
  - add functions to utils to generate these paths from the corresponding id
  - refactor some child endpoints
  - remove `filename` and `approved` fields from MilestoneImage
    - filename is redundant
    - I think approved can be done better with a separate model for user-provided images waiting for admin approval (not in this PR)
  - remove default child image, instead return 404
    - resolves #169
  - rename databse from milestones.db -> mondey.db
- frontend
  - update image urls
  - ChildrenGallery
    - don't display error if child doesn't have an image
    - use URL.createObjectURL instead of FileReader


MaHaWo (approved):
lgtm

PR #177, opened 2024-11-19
Optimise images when they are uploaded and save in webp format

### LINKED ISSUE #166: Optimize image sizes
It would be good at some point to optimize the image sizes, since using the site will involving downloading a lot of images.

- can be done by admins before uploading the images (no effort on our side, but inconvenient for the admins)
- or as a one-off / sporadically using a script on the server (limited effort on our side, reasonable convenience for admins)
- or it could be integrated into the backend as a utility function that runs automatically (larger initial effort)

lkeegan:
- backend
  - open uploaded images using pillow
  - resize if width > 1024
  - export to webp format using pywebp
  - add some png & jpeg files to the tests
- frontend-admin
  - allow png uploads as well as jpg
- frontend
  - update image urls
- resolves #166

### DIFF
--- a/frontend/src/lib/admin.svelte.ts
+++ b/frontend/src/lib/admin.svelte.ts
@@ -21,7 +21,7 @@ export async function refreshMilestoneGroups() {
 }
 
 export function milestoneGroupImageUrl(id: number) {
-	return `${import.meta.env.VITE_MONDEY_API_URL}/static/mg/${id}.jpg`;
+	return `${import.meta.env.VITE_MONDEY_API_URL}/static/mg/${id}.webp`;
 }
 
 export async function refreshUserQuestions() {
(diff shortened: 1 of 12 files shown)


MaHaWo (commented):
Nice improvement. Code looks good, but it seems it breaks the childrengallery: 
- when I add an image for a child, it cannot be retrieved: 
```
Error in server request
Failed to execute 'createObjectURL' on 'URL': Overload resolution failed.
```
should we remove the image display for children altogether first? From the discussion about the default image I gathered that this is not desired anymore? Or did that just apply to the default image?

lkeegan:
> Nice improvement. Code looks good, but it seems it breaks the childrengallery:
> 
> * when I add an image for a child, it cannot be retrieved:
> 
> ```
> Error in server request
> Failed to execute 'createObjectURL' on 'URL': Overload resolution failed.
> ```
> 
good catch - should be fixed by 8e56030

> should we remove the image display for children altogether first? From the discussion about the default image I gathered that this is not desired anymore? Or did that just apply to the default image?

I understood that it only applied to the default image, which was not wanted, but we should still show an image if the user provides one.

PR #178, opened 2024-11-20
Limit allowed file size to 2MB for image uploads

### LINKED ISSUE #165: Image upload file size limits
To avoid the site becoming very slow to load for users with slower connections, we should enforce some limits on the size of image files that can be uploaded:

- on the frontend side, e.g. 500kb
  - should apply to admin image uploads as well as user uploads
- on the server side, e.g. 1MB
  - backup limit in case a large file gets through the frontend

lkeegan:
- admin-frontend
  - limit upload image file sizes to 2 MB for Milestone and MilestoneGroup images
    - display warning modal and ignore files if too large
- frontend
  - add WarningModal and ImageFileUpload components
- deployment
  - set client_max_body_size to 2MB in nginx config
- backend
  - remove exif orientation tag if present to avoid unexpected rotation of uploaded images in some cases
  - use ImageOps.contain instead of resizing the image
- resolves #165

### DIFF
--- a/frontend/src/lib/components/WarningModal.svelte
+++ b/frontend/src/lib/components/WarningModal.svelte
@@ -9,7 +9,7 @@ let { open = $bindable(false), text }: { open: boolean; text: string } =
 	$props();
 </script>
 
-<Modal bind:open size="xs" autoclose>
+<Modal bind:open size="xs" autoclose outsideclose>
 	<div class="text-center">
 		<ExclamationCircleOutline class="mx-auto mb-4 h-12 w-12 text-gray-400 dark:text-gray-200" />
 		<h3 class="mb-5 text-lg font-normal text-gray-500 dark:text-gray-400">
--- a/frontend/src/lib/components/WarningModal.svelte
+++ b/frontend/src/lib/components/WarningModal.svelte
@@ -1,0 +1,20 @@
+<svelte:options runes={true} />
+
+<script lang="ts">
+import { Button, Modal } from "flowbite-svelte";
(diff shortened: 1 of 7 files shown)


frontend/src/lib/components/WarningModal.svelte
```diff
@@ -0,0 +1,20 @@
...
+<script lang="ts">
+import { Button, Modal } from "flowbite-svelte";
+import ExclamationCircleOutline from "flowbite-svelte-icons/ExclamationCircleOutline.svelte";
+import { _ } from "svelte-i18n";
+
+let { open = $bindable(false), text }: { open: boolean; text: string } =
+	$props();
+</script>
+
+<Modal bind:open size="xs" autoclose>
```
MaHaWo, line 12: would it be possible to make this such that it leaves the parent modal open? currently it seems that,e.g., when I upload an image that is too large in the milestonegroup it throws me back to the milestonegroup overview and I have to open the edit element for the milestonegroup again to try another image.

MaHaWo (approved):
lgtm. One minor comment wrt  usability. I'm not sure if we should fix such things as they appear, or if we should collect them in a meta issue and work through them in a more focused way?

frontend/src/lib/components/WarningModal.svelte
lkeegan: good catch! I guess I never clicked the Verstanden button, as if you click on the x in the top right of the modal this doesn't happen.
And yes I think this PR is the right place to fix this issue - but we can and should do some usability testing at some point looking for minor niggles like this

lkeegan: should be fixed by 4440830

PR #190, opened 2024-11-26
Make flaky test robust

lkeegan:
- expected age can vary by +/- 1 month depending on exact date when test runs
- this is due to making the answer session 30 days in the past, which is not always the previous month
- allow for this in the test so the test is valid regardless of the current date!

### DIFF
--- a/mondey_backend/pyproject.toml
+++ b/mondey_backend/pyproject.toml
@@ -21,6 +21,8 @@ dependencies = [
     "click",
     "pillow",
     "webp",
+    "python-dateutil",
+    "types-python-dateutil"
 ]
 dynamic = ["version"]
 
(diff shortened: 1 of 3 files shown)


MaHaWo:
ok, I just now noticed this and used [dateutil](https://dateutil.readthedocs.io/en/latest/) to fix this in `conftest.py` file like this: 

```python 

@pytest.fixture
def children():
    today = datetime.datetime.today()
    nine_months_ago = today - relativedelta(months=9)
    twenty_months_ago = today - relativedelta(months=20)
```

what should we go with?

lkeegan:
> ok, I just now noticed this and used [dateutil](https://dateutil.readthedocs.io/en/latest/) to fix this in `conftest.py` file like this:
> 
> ```python
> @pytest.fixture
> def children():
>     today = datetime.datetime.today()
>     nine_months_ago = today - relativedelta(months=9)
>     twenty_months_ago = today - relativedelta(months=20)
> ```
> 
> what should we go with?

nice - yours looks better to me

PR #191, opened 2024-11-27
Fix the set of milestones when a milestone answer session is created

### LINKED ISSUE #189: a MilestoneAnswerSession should have a fixed set of milestone groups and milestones
Currently we construct the appropriate milestone groups and milestones for a child on the fly based on their age, in `get_milestone_groups`

We then only create a `MilestoneAnswer`, and if needed a new `MilestoneAnswerSession`, when an answer is submitted.

This is robust to changes in which milestones are included by get_milestone_groups, but it is not ideal for calculating the traffic light feedback, in particular the age-appropiate milestones may have changed, and we also need to know if there were milestones that were not answered.

For calculating the feedback, it would be better if all the MilestoneAnswers were created when a MilestoneAnswerSession is created, and for the milestone group to be easily determined from a MilestoneAnswer.

Todo:

- get_milestone_groups
  - should call get_or_create_current_milestone_answer_session
- get_or_create_current_milestone_answer_session
  - should create a MilestoneAnswer for each milestone in the session
  - with `none` or equivalent as the answer
- get_current_milestone_answer_session
  - maybe combine with get milestone groups?
  - it could return e.g. `{milestones: ..., answers: ...}`, now that both are created together
- MilestoneAnswer
  - should have a `milestone_group_id` field for convenience when calculating feedback / stats
  - needs a `none` value for the answer (this could be e.g. -1)
- frontend milestone component
  - update logic accordingly

lkeegan:
- MilestoneAnswer
  - An answer is now created for every milestone when an answer session is created
  - The initial value of the answer is `-1` which indicates the user has not yet submitted an answer for this milestone
  - Add `milestone_group_id` field for convenience when calculating statistics / feedback
- get_milestone_groups
  - Now returns the milestones according to the MilestoneAnswers for the current session
  - This means the set of milestones is fixed and doesn't change even if a milestone expected_age changes such that it would no longer be selected for that session
- update_milestone_answer
  - no longer creates an answer session if there is no existing unexpired one
  - either updates the existing answer for the supplied session, or returns 404 or 401
- Milestone component
  - all answers now exist, replace check for existence with check for `-1` to determine if the answer has already been answered
- add types-python-dateutil to mypy pre-commit hook dependencies
  - remove out-dated comment & type ignores from related code
- resolves #189

### DIFF
--- a/mondey_backend/src/mondey_backend/routers/milestones.py
+++ b/mondey_backend/src/mondey_backend/routers/milestones.py
@@ -16,8 +16,8 @@ from ..models.milestones import MilestonePublic
 from ..models.milestones import SubmittedMilestoneImage
 from .utils import add
 from .utils import get
-from .utils import get_child_age_in_months
 from .utils import get_db_child
+from .utils import get_or_create_current_milestone_answer_session
 from .utils import submitted_milestone_image_path
 from .utils import write_image_file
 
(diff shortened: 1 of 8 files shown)


lkeegan:
(to test this locally it would be good to first do `DROP TABLE milestoneanswer;` and `DROP TABLE milestoneanswersession;`)

mondey_backend/src/mondey_backend/routers/milestones.py
```diff
@@ -52,25 +52,18 @@ def get_milestone_groups(
...
     ):
-        delta_months = 6
         child = get_db_child(session, current_active_user, child_id)
-
-        child_age_months = get_child_age_in_months(child)
+        milestone_answer_session = get_or_create_current_milestone_answer_session(
+            session, current_active_user, child
+        )
+        milestone_ids = list(milestone_answer_session.answers.keys())
+        print("milestone_ids", milestone_ids)
```
MaHaWo, line 60: is this useful for upcoming changes or is this debug output that slipped through the net?

MaHaWo (approved):
lgtm

mondey_backend/src/mondey_backend/routers/milestones.py
lkeegan: a print statement is always the latter, thanks :-)

PR #193, opened 2024-11-27
add a carousel element with bookmark-tabs at the front page

Majapur1:
- Add a new element containing information about monitoring areas in MONDEY;
- Implement an auto-changing timer for the tabs;

- Update of /de.json file.

### DIFF
--- a/frontend/src/lib/components/FrontPageBookmarks.svelte
+++ b/frontend/src/lib/components/FrontPageBookmarks.svelte
@@ -1,28 +1,28 @@
 <script>
-    export let bookmarks = [];
-    import { _ } from "svelte-i18n";
+export let bookmarks = [];
+import { _ } from "svelte-i18n";
 
-    let activeIndex = 0;
-    let interval;
+let activeIndex = 0;
+let interval;
 
-    function changeTab() {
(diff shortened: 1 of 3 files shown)


Majapur1:
@lkeegan I just realized that I could use the Flowbite Svelte "Tabs" element instead of implementing it this way. 

Should I do it one more time with a Flowbite version?

lkeegan:
@Majapur1 sorry I won't have time to review this until next week - but yes using the flowbite-svelte tab component for this sounds like a good idea.

lkeegan (approved):
@Majapur1 thanks, this looks good - and probably makes sense not to use the flowbite tab if you want to keep your vertical tab layout on mobile.

Also something to maybe consider for this kind of thing (if you didn't already) is to use a carousel component, e.g.
- https://flowbite-svelte.com/docs/components/carousel
- https://vadimkorr.github.io/svelte-carousel/

But lgtm!

Majapur1:
@lkeegan thanks!

Probably I've used the wrong name for this element. At the end it's more like "Tabs" than "Carousel" (without pictures and icons), but thanks for these links!

Yes, I want to keep it vertical on mobile, what could be a problem with pre-added style-settings of Flowbite elements.

PR #194, opened 2024-11-28
Implement fixed age intervals

MaHaWo:
### first version (obsolete)
- done in preparation to the statistics computation things in the feedback backend
- no fixed milestone `timedelta` anymore (that we had set to 6 months), but a configurable interval per milestone 
- backend endpoints for adding, removing and modifying age intervals
  - I'm not entirely sure if allowing modification is a good idea as it will invalidate previous analysis/feedbacks 
- milestones have an new `ageinterval` member that tells to which interval they belong. 
  - there would be the possibility to compute this interval automatically if the data is good enough, but I'm not sure we can do this reliably, and would prefer if they would do that on their side of things and then just put in the numbers instead of giving us some algorithm for it that we have to implement and maintain and be responsible for
- new tab in admin frontend interface for the age intervals 
- new element in the milestone frontend to select it. 
- adjusted the database query that pulls out the relevant answers for the milestoneanswersession based on time. 
- vast majority of the indicated changes come from the regenerated `openapi.json` file. 

## revised version 
- new elements `expected_age_months_minus` and `expected_age_months_plus` in milestone database models.
- these are relative to `expected_age_months`
- adjusted the database query that pulls out the relevant answers for the milestoneanswersession based on time. 
- age interval for a  `milestone = [expected_age_months-expected_age_months_minus, expected_age_months + expected_age_months_plus] `
- add three-way range slider frontend component and replace the slider for `expected_age_months` in `EditMiletoneModal` with that to edit the age interval for a milestone in one go. 
   - styling was surprisingly difficult to get consistent between firefox and chromium, and it's still a bit off :(


lkeegan:
I guess it's all implemented in other PRs by now? So yes, let's close the PR but keep the branch in case

MaHaWo:
no, it's not. in fact, I went to some lengths to not pollute other PRs with this....

PR #197, opened 2024-12-05
Add researcher codes

### LINKED ISSUE #185: Research users should have a code
This can be added to the `User` object, or stored elsewhere.

This code is then entered by users when they sign up, which provides their data to that researcher

lkeegan:
- backend
  - add ResearchGroup model with 6-digit id code for users to use when signing up
  - add checkdigit dependency to add checksum to code
  - add full_data_access flag to users if they should have research access to all data
- admin frontend users tab
  - add reseach_group_id & full_data_access to table
  - add button to add create a research code for a researcher
- frontend
  - add ResearchCodeInput input which checks the entered code checksum is valid
  - add optional code input to user signup, remove role selection
  - make url for signup /signup
  - add optional code e.g. /signup/123451 to allow researchers to send sign-up links to their users with code pre-filled
  - add cdigit dependency to validate codes
- resolves #185


MaHaWo (approved):
code looks good. 
- Wrt the optional code in the signup link: Could this functionality potentially be misused if the link is accidentally sent to a wrong person for example? Are these signup links valid forever? 
- does the checksum mechanics have additional use cases (in the research data download perhaps?)

lkeegan:
> code looks good.
> 
> * Wrt the optional code in the signup link: Could this functionality potentially be misused if the link is accidentally sent to a wrong person for example? Are these signup links valid forever?
> * does the checksum mechanics have additional use cases (in the research data download perhaps?)

yes the code is valid forever, and associates the user with the researcher. But what we should do is check that the provided code is an existing research code & if not maybe just remove it, to avoid the unlikely scenario where a user signs up with a random valid code, and later a researcher is created with by coincidence exactly this code

PR #198, opened 2024-12-05
Add email verification

### LINKED ISSUE #109: add email to backend

lkeegan:
- docker-compose
  - add postfix image
- backend
  - add SMTP_HOST to settings
  - send user an email with a verification link when they sign up
- frontend
  - add /verify route
- resolves #109

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -28,9 +28,7 @@ services:
     volumes:
       - ${MONDEY_SSL_CERT:-./cert.pem}:/mondey_ssl_cert.pem
       - ${MONDEY_SSL_KEY:-./key.pem}:/mondey_ssl_key.pem
-#  email:
-#    image: "boky/postfix"
-#    environment:
-#      - ALLOW_EMPTY_SENDER_DOMAINS="true"
-#    networks:
-#      - mondey-network
+  email:
+    image: "boky/postfix"
+    environment:
+      - ALLOW_EMPTY_SENDER_DOMAINS="true"
--- a/mondey_backend/src/mondey_backend/users.py
+++ b/mondey_backend/src/mondey_backend/users.py
@@ -1,8 +1,8 @@
-# TODO: 17th Oct. 2024: remove the artificial verification setting again as soon as
-# the email verification server has been implemented. See 'README' block @ line 33f
-
 from __future__ import annotations
 
+import logging
+import smtplib
+from email.message import EmailMessage
 from typing import Annotated
 
 from fastapi import Depends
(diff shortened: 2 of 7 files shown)


mondey_backend/src/mondey_backend/users.py
```diff
@@ -18,29 +18,30 @@
...
 from .settings import app_settings
 
 
+def send_email_validation_link(email: str, token: str) -> None:
+    msg = EmailMessage()
+    msg["From"] = "no-reply@mondey.lkeegan.dev"
+    msg["To"] = email
+    msg["Subject"] = "MONDEY-Konto aktivieren"
+    msg.set_content(
+        f"Bitte klicken Sie hier, um Ihr MONDEY-Konto zu aktivieren:\n\nhttps://mondey.lkeegan.dev/verify/{token}"
```
MaHaWo, line 32: should this be localized in some way or contain an english translation?

docker-compose.yml
```diff
@@ -28,9 +28,7 @@ services:
...
-#  email:
-#    image: "boky/postfix"
-#    environment:
-#      - ALLOW_EMPTY_SENDER_DOMAINS="true"
-#    networks:
-#      - mondey-network
+  email:
+    image: "boky/postfix"
+    environment:
+      - ALLOW_EMPTY_SENDER_DOMAINS="true"
```
MaHaWo, line 34: I looked this up [here]( https://github.com/bokysan/docker-postfix/blob/master/README.md#allowed_sender_domains-and-allow_empty_sender_domains) ) but could only find documentation in conjunction with `ALLOWED_SENDER_DOMAINS`. Is this temporary?

mondey_backend/tests/routers/test_auth.py
```diff
@@ -0,0 +1,41 @@
...
+        SMTPMock.last_message = msg
+
+
+@pytest.fixture
+def smtp_mock(monkeypatch: pytest.MonkeyPatch):
+    monkeypatch.setattr(smtplib, "SMTP", SMTPMock)
+    return SMTPMock
+
+
+def test_register_new_user(public_client: TestClient, smtp_mock: SMTPMock):
```
MaHaWo, line 30: do we need a test for failed verification?

MaHaWo (approved):
lgtm, thanks for doing this. Have 3 little questions regarding code which are more for clarification. I didn´t get it to work locally though to try it out.

*Edit:* works with iwr address, but not with google?

docker-compose.yml
lkeegan: `ALLOWED_SENDER_DOMAINS` is I think by default empty, so setting `ALLOW_EMPTY_SENDER_DOMAINS` tells postfix to allow sending emails from any domain

mondey_backend/src/mondey_backend/users.py
lkeegan: I guess so, I'd go with including an english translation in the email to keep this simple

lkeegan:
> lgtm, thanks for doing this. Have 3 little questions regarding code which are more for clarification. I didn´t get it to work locally though to try it out.
> 
> _Edit:_ works with iwr address, but not with google? I also got a 404 when clicking on the link provided, but could log in anyway after that?

Yes google are quite aggressive in blocking unknown / potentially spam domains. We'll have to see if it works when sending from mondey.de or if we need an email account or maybe sign up for a mail sending service.

The link should work if you replace mondey.lkeegan.dev with localhost in the url when running locally?

mondey_backend/tests/routers/test_auth.py
lkeegan: yes true, will add one

PR #208, opened 2024-12-13
add a new accordion element on the front page 

Majapur1:
### New element on the front page

- add an accordion element with an information about the project with a function to parsing markdown from `de.json` (that helps to avoid a huge amount of lines in `de.json` for just one markdowned description);
- update of `de.json` file;
- add some extra additional colors into `tailwind.config.ts`.


lkeegan (commented):
thanks @Majapur1 - as we just discussed, I'd suggest that for now
- you directly put the HTML in de.json
- and remove the parseMarkdown function
I'll look into #210 and #211 to follow up on this

PR #213, opened 2024-12-17
Refactor i18n implementation

### LINKED ISSUE #211: Allow admins to edit the german translations
Right now de.json is hard coded, some technical issues to resolve, but this should ideally also be editable by the admins from the admin interface

lkeegan:
- main improvements
  - all language texts can now be edited from the admin interface
  - in frontend code we now have autocomplete for translation ids
- change in use
  - i18n-svelte -> lib/i18n.svelte
  - `$locale` -> `i18n.locale`
  - `$locales` -> `i18n.locales`
  - `$_('a.b')` -> `i18n.tr.a.b`
  - to add a text, add the id to lib/translations.ts
- code changes
  - remove i18n-svelte library
  - remove locales/de.json: this is now included in the static folder of the deployed website, not part of the frontend source code anymore
  - use camel case for all translation ids for consistency and to allow use without quotes
  - de is now also editable in admin interface
- resolves #211


MaHaWo (approved):
nice addition, looks good. The only question I have is if there would be a way to have defaults for, e.g., the names of the buttons in the profile popover or similar functional elements. Someone who is not familiar with the code might be a little confused by the amount of names and labels and doesn´t know what to provide there upon initial deployment.

lkeegan:
> nice addition, looks good. The only question I have is if there would be a way to have defaults for, e.g., the names of the buttons in the profile popover or similar functional elements. Someone who is not familiar with the code might be a little confused by the amount of names and labels and doesn´t know what to provide there upon initial deployment.

Yes I was also thinking about that and I agree - I guess the most developer-friendly option would be to put the actual german text in https://github.com/ssciwr/mondey/pull/213/files#diff-63f05097c4a548b2bc2106f45ca89b7986d834aaddc361c23ea8b1850c2e59b6 instead of empty strings, even if this wastes some bandwidth.

The alternative would be to ship a de.json as part of the source code but not part of the website, and modify the local docker / testing setup to serve that at runtime, but don't think it's worth the extra complications just to avoid downloading some redundant data, at least for now.

PR #214, opened 2024-12-17
Add trafficlight feedback frontend

MaHaWo:
- refactor the old feedback frontend from #183 into the latest available design from @Majapur1 
- add in a max number of answersessions that are displayed at once (currently 6) 
- fix styling to work better with mobile screen sizes
- add better explanations

PR #218, opened 2024-12-20
Add color to children

### LINKED ISSUE #154: Add colour to children
Children should have a colour that the user can choose, which is used as the background

MaHaWo:
- add background color to children card display in `ChildrenGallery` 
- child database table has a `color` column now 
- add color input element to the `ChildrenRegistration` component 
- use color with and without image, since image gets removed on screens with small width 
- adjust styles and tests as necessary
- same flaky test syndrome as the other PR, fixed in #214 
fixes #154


lkeegan (approved):
needs rebasing on main, otherwise lgtm

PR #232, opened 2025-03-05
Rework feedback page

### LINKED ISSUE #227: New client suggestions on the feedback page
- This issue is a part of #223 and #222 

Below is a series of client comments on the appearance of this page:

- we need to add the title; 
- the description should be changed as in the example (volume reduction); 
- all texts (descriptions, details under the button and legend) need to match the latest version of the client (I'll attach it below);
- the legend should stay open all the time without wrapping (we can reduce the scrolling time with a shorter description);
- get rid of the buttons "Feedback zu vorherigen Beobachtungszeiträumen anzeigen", "Mehr Informationen zur Erstellung der Bewertung" and blue arrows around the previous periods. It can just be there all the time;
- add the button "Mehr Info zur Bewertung" at the same level as the second title (I'll attach the new text of it below);
- the customer's request is that the primary color (blue) be used only for milestones. The accordion headers need to be blue. For other elements on the site we can use the `"additional-color"` options. They are already added to `tailwind.config.ts`;
- the feedback for the overall score should be presented before the Individual development areas;
- can we have 3 or 2 columns of milestone groups on wide screens;
- The user should only be able to see the problematic milestones under a milestones group (green ones should be invisible). The background of a problematic milestone on opening is (yellow) color `#f3ddb0`.

Image

MaHaWo:
- [x] we need to add the title;
  - [x] the description should be changed as in the example (volume reduction);
  - [x] all texts (descriptions, details under the button and legend) need to match the latest version of the client (I'll attach it below);
  - [x] the legend should stay open all the time without wrapping (we can reduce the scrolling time with a shorter description);
  - [x] get rid of the buttons "Feedback zu vorherigen Beobachtungszeiträumen anzeigen", "Mehr Informationen zur Erstellung der Bewertung" and blue arrows around the previous periods. It can just be there all the time;
  - [x] add the button "Mehr Info zur Bewertung" at the same level as the second title (I'll attach the new text of it below);
  - [x] the customer's request is that the primary color (blue) be used only for milestones. The accordion headers need to be blue. For other elements on the site we can use the "additional-color" options. They are already added to tailwind.config.ts;
 - [x] the feedback for the overall score should be presented before the Individual development areas;
 - [x] can we have 3 or 2 columns of milestone groups on wide screens; 
 - [x] The user should only be able to see the problematic milestones under a milestones group (green ones should be invisible). The background of a problematic milestone on opening is (yellow) color #f3ddb0.

closes #227


MaHaWo:
- [x] we need to add the title;
  - [x] the description should be changed as in the example (volume reduction);
  - [x] all texts (descriptions, details under the button and legend) need to match the latest version of the client (I'll attach it below);
  - [x] the legend should stay open all the time without wrapping (we can reduce the scrolling time with a shorter description);
  - [x] get rid of the buttons "Feedback zu vorherigen Beobachtungszeiträumen anzeigen", "Mehr Informationen zur Erstellung der Bewertung" and blue arrows around the previous periods. It can just be there all the time;
  - [x] add the button "Mehr Info zur Bewertung" at the same level as the second title (I'll attach the new text of it below);
  - [ ] the customer's request is that the primary color (blue) be used only for milestones. The accordion headers need to be blue. For other elements on the site we can use the "additional-color" options. They are already added to tailwind.config.ts;
 - [x] the feedback for the overall score should be presented before the Individual development areas;
 - [x] can we have 3 or 2 columns of milestone groups on wide screens; 
 - [x] The user should only be able to see the problematic milestones under a milestones group (green ones should be invisible). The background of a problematic milestone on opening is (yellow) color #f3ddb0.

MaHaWo:
@Majapur1 
in the vid below you can see the current version of the feedback page  this branch implements, in case you can´t get it to work locally: 

[feedback_page.webm](https://github.com/user-attachments/assets/6d2442d7-90ce-4e6c-a024-3cf0f4a55a11)

Majapur1:
@MaHaWo It looks good, thanks for the videos!

Regarding your reasonable question about the yellow milestones under the green groups — I have an answer. We need to add information (a line of text) about this particular possibility to inform users. I will get back to you with ideas on where to place it and some comments on the latest version of the page.

frontend/src/lib/components/ChildrenFeedback.svelte
```diff
@@ -97,6 +128,16 @@ const breadcrumbdata: any[] = [
...
+// TODO: for performance reasons, we could consider moving some of these functions to the backend,
+// but this can be done once performance becomes a problem.
+
+/**
+ * Get the answersessions for the child we are concerned with from the database.
+ * @summary Get the answersessions for the child we are concerned with from the database, set the sessionkeys as a sorted and reversed list of keys that we get from the response, and set the relevant_sessionkeys to be the interval of sessionkeys that we want to show.
+ * @return {Promise<void>} nothing
+ */
 async function loadAnswersessions(): Promise<void> {
 	user.load;
```
lkeegan, line 142: should this be `await user.load()`?

frontend/src/lib/components/ChildrenFeedback.svelte
```diff
@@ -175,41 +226,27 @@ async function loadDetailedFeedback(relevant: number[]): Promise<void> {
...
 
-	await loadSummaryFeedback(relevant_sessionkeys);
-	await loadDetailedFeedback(relevant_sessionkeys);
+		detailed[Number(aid)] = res;
+	}
 }
 
+/**
+ * Generate a printable report from the feedback by concatenating all the feedbacks into a single formatted string.
+ * @summary If the description is long, write your summary here. Otherwise, feel free to remove this.
```
lkeegan, line 247:
```suggestion
```

frontend/src/lib/components/ChildrenFeedback.svelte
```diff
@@ -263,6 +305,11 @@ function printReport() {
 	printWindow.print();
 }
 
+/**
+ * Function for formatting a given datetime string into a more readable format.
+ * @param {string} date - The date to format.
+ * @return {string} The formatted date: day - month - year
+ */
 function formatDate(date: string): string {
```
lkeegan, line 313: for consistency we should probably use this everywhere in the report, e.g. replace the `new Date().toLocaleDateString()` above

frontend/src/lib/components/ChildrenFeedback.svelte
```diff
@@ -39,24 +36,56 @@ import {
...
+// helpers
+const breakpoints = {
+	sm: 640,
+	md: 768,
+	lg: 1024,
+	xl: 1280,
+	"2xl": 1536,
+};
+
+// TODO: make this reactive such that the page reloads when the width changes
```
lkeegan, line 60: I'm not sure we want to reactively do a bunch of API calls if the screen width changes?
Personally I would display some fixed number of answer sessions initially, with an "older" or "more" button the user can click to load more if desired, rather than hard-coding the number of answer sessions to display based on the screen width.
Also it might make sense to only fetch feedback for an answer session when it needs to be displayed instead of loading them all before displaying anything (depending on how long this ends up taking with real data)
But I'm fine with this as is for now, maybe we can come back to it once the website is complete and we have some experience trying to use it.

lkeegan (approved):
thanks lgtm, some minor comments

frontend/src/lib/components/ChildrenFeedback.svelte
MaHaWo: yes that's true. I would have kept the back/forward buttons, but the psychologists don´t want them (see discussion in the linked issue -> "get rid of the buttons "Feedback zu vorherigen Beobachtungszeiträumen anzeigen", "Mehr Informationen zur Erstellung der Bewertung" and blue arrows around the previous periods. It can just be there all the time". 
The ensuing discussion between Igor and myself ended with the history being available in the printout but not being displayed on the page.  Therefore I opted for the hardcoded solution. 

`Also it might make sense to only fetch feedback for an answer session when it needs to be displayed instead of loading them all before displaying anything (depending on how long this ends up taking with real data)`

good point as well, I feel like we have hit upon this at some point already. We could subsume it under a `performance improvements` issue, or look into it here right away to not get it buried under other stuff?

lkeegan: > I'm not sure either. I would much rather have kept the back/forward buttons, but the psychologists don´t want them (see discussion in the linked issue -> `get rid of the buttons "Feedback zu vorherigen Beobachtungszeiträumen anzeigen", "Mehr Informationen zur Erstellung der Bewertung" and blue arrows around the previous periods. It can just be there all the time`. The ensuing discussion between Igor and myself (it cannot be there all the time because history gets too big while some access to it should be provided regardless) ended with the compromise that the history should be available in the printout but not on the page. Therefore I opted for the hardcoded number of answersessions adjusted to screen size.
> 
Ok that makes sense, thanks, I missed that discussion

> `Also it might make sense to only fetch feedback for an answer session when it needs to be displayed instead of loading them all before displaying anything (depending on how long this ends up taking with real data)`
> 
> good point as well, I feel like we have hit upon this at some point already. We could subsume it under a `performance improvements` issue, or look into it here right away to not get it buried under other stuff? I think the current solution with everything being loaded at the start is an artifact of when we could flick through the entire history with the arrow buttons, which I didn't want to have loading delays in between. That was payed for by a longer loading time for the page itself. On smaller devices that might become an issue at some point though as history gets longer of course. I ignored this issue at the time, relying on future experience to decide on it....

If you don't mind looking into this now, what do you think about this:

- we load all answer sessions initially, and as currently choose how many to display based on the screen size
  - but we don't load the feedback initially (apart from the latest one which is displayed)
- when the user clicks on a session we load the feedback if it is not present
- similarly when they click on print we first load any missing feedback

This way we shouldn't have any performance issues even if there are loads of answer sessions, and the number of displayed answer sessions could also be reactive, since changing this no longer involves any API calls?

MaHaWo: I noticed a bug in the printout function however, which would only print the currently displayed answersessions (and that didn´t come up because I didn´t have enough yet).

MaHaWo: >>> 
If you don't mind looking into this now, what do you think about this:

we load all answer sessions initially, and as currently choose how many to display based on the screen size
but we don't load the feedback initially (apart from the latest one which is displayed)
when the user clicks on a session we load the feedback if it is not present
similarly when they click on print we first load any missing feedback
>>> 

that sounds like a good plan. Will try to put this in right away to be done with this PR.

MaHaWo: ok, so, as usual this took way longer than expected because I messed up an promise-await chain initially and then found that the endpoint for getting expired milestoneanswersesssions didn´t make use of the correct flags yet and lacked a test as well. Do you want to have another look @lkeegan ?

frontend/src/lib/components/ChildrenFeedback.svelte
```diff
@@ -130,24 +149,23 @@ const breadcrumbdata: any[] = [
...
-
 /**
  * Get the answersessions for the child we are concerned with from the database.
  * @summary Get the answersessions for the child we are concerned with from the database, set the sessionkeys as a sorted and reversed list of keys that we get from the response, and set the relevant_sessionkeys to be the interval of sessionkeys that we want to show.
  * @return {Promise<void>} nothing
  */
 async function loadAnswersessions(): Promise<void> {
-	user.load;
+	if (!user.data) {
+		await user.load;
```
lkeegan, line 159:
```suggestion
		await user.load();
```

mondey_backend/src/mondey_backend/routers/users.py
```diff
@@ -246,9 +245,10 @@ def get_expired_milestone_answer_sessions(
                 select(MilestoneAnswerSession).where(
                     (col(MilestoneAnswerSession.user_id) == current_active_user.id)
                     & (col(MilestoneAnswerSession.child_id) == child_id)
+                    & col(MilestoneAnswerSession.expired)
+                    & col(MilestoneAnswerSession.included_in_statistics)
```
lkeegan, line 249: Why the name change / requirement for `included_in_statistics` to be true here?
I don't think it's necessary to be able to construct feedback for the answer session?

mondey_backend/src/mondey_backend/routers/users.py
```diff
@@ -246,9 +245,10 @@ def get_expired_milestone_answer_sessions(
                 select(MilestoneAnswerSession).where(
                     (col(MilestoneAnswerSession.user_id) == current_active_user.id)
                     & (col(MilestoneAnswerSession.child_id) == child_id)
+                    & col(MilestoneAnswerSession.expired)
```
lkeegan, line 248: nice catch

lkeegan (approved):
thanks - lgtm apart from the included_in_statistics comment

mondey_backend/src/mondey_backend/routers/users.py
MaHaWo: well, the function no longer just returns expired milestonesessions, and it's only used on the feedback page.

PR #233, opened 2025-03-07
Unify navbars

MaHaWo:
- use sidebar instead of tabs for userland navigation 
- make the sidebar into a drawer with a button to show it when on smaller screens 
- make breadcrumbs more useful by adding starting page of navigation 
- add dummy settings element 
- make the app use the darkmode setting of the user's system
- delete obsolete files 
- update packages
- removed redundancies between navbar and footer

open still for suggestions:
- stop navbar from collapsing too early 
- what to do with it other than having a logo and the locale chooser?

separate PR:
- tests (separate PR since almost nothing has behavioral tests yet)


jmsssc:
I am still testing the mobile version of the menus, by the way. But the PC version is ready now with the moved content/buttons.

MaHaWo:
@jmsssc is this ready for review ?

jmsssc:
> @jmsssc is this ready for review ?

Yep, this is ready

frontend/src/lib/components/DarkModeChooser.svelte
```diff
@@ -0,0 +1,12 @@
+<script>
+import FunctionalIcon from "$lib/components/Navigation/FunctionalIcon.svelte";
+import { DarkMode } from "flowbite-svelte";
+import { MoonSolid, SunSolid } from "flowbite-svelte-icons";
+</script>
+
+<FunctionalIcon tooltip={'Darkmode ein- oder ausschalten'}>
```
lkeegan, line 7: this german tooltip text can be added somewhere in https://github.com/ssciwr/mondey/blob/main/frontend/src/lib/translations.ts
and referenced here, e.g. something like
```
...
...
tooltip={i18n.tr.frontpage.toggleDarkmode}
```

then in the admin interface in the deployed version translations can be added for other languages

lkeegan (commented):
Thanks, this looks good!

A couple of issues I noticed:

- when logged in, trying to change the locale doesn't change the locale, and appears to also reload the "kinder" page in de locale?
- when logged in with a mobile size screen, sometimes clicking on the locale chooser opens the menu at the top of the screen instead of where I clicked
  - could this be because in LocaleChooser the Dropdown has `triggeredBy="#id"`, with the same id for both of our instances of LocaleChooser?
- "einstellung" link doesn't seem to work
- (very minor) the hamburger in mobile view & dark mode is a bit bigger than the top grey area - and maybe would benefit from being white-ish in dark mode to make it more obvious? I guess ideally it would be the same icon that is there in mobile view when the user is not logged in?
image

frontend/src/lib/components/DarkModeChooser.svelte
jmsssc: Thank you for the review @lkeegan 

Ah good catch sure absolutely the Hamburger icon is an issue on dark mode, that's wrong. I made the icon bigger but I can return it to the same size as the non-logged in icon (lg rather than xl) and make it turn white in dark mode. And maybe remove the background color of it's container if it messes with the UI. I will make the tooltip a translation key too.

On the other couple of things: The LocaleChooser I thought wasn't functional yet throughout, but absolutely I will look at why it might not work now and get it working on this branch. Einstellung is the same, it was like that so I did not change it or think anything was awry as I didn't realize it worked differently on main. I will look at code in main where I see it allows you to change your password and set locale, and get it working on this branch.

By the way, I've had an issue with my PC freezing after lunch and can't get past boot, so I'm working from my laptop, as Stefan/the guy on this floor aren't in. If I can't access git from this computer to push, I'll be able to send myself patch files to push them tomorrow.

jmsssc:
In addition to what Liam noted, I think the static mobile menu could be improved, especially in dark mode, in terms of padding around the icon buttons and making them consistent. Possibly also changing it to be a sidebar like the logged in menu on mobile (if this does not entail too much code change/duplicate code) because the sidebar feels a lot nicer on mobile:

Any thoughts on this?

image

jmsssc:
> @jmsssc is this ready to review/merge?

Yes, it should be ready now (just pushed a commit - to remove some old console.logs)

I'm still not sure about the Settings page being right by the way (I think it might be my translations config/cache, does it look right to you?)

lkeegan:
> > @jmsssc is this ready to review/merge?
> 
> Yes, it should be ready now (just pushed a commit - to remove some old console.logs)
> 
> I'm still not sure about the Settings page being right by the way (I think it might be my translations config/cache, does it look right to you?)

Thanks - settings page seems ok to me after I added some translations in the admin/Übersetzungen section & clicked save

frontend/src/lib/components/UserLandingPage.svelte
```diff
@@ -1,98 +1,193 @@
...
+						currentChild.data = null;
+						hideDrawer = true;
+						goto(`/${base}`);
+					}
+				}}>
+					<svelte:fragment slot="icon">
+						<ArrowRightToBracketOutline size="lg" />
+					</svelte:fragment>
+				</SidebarItem>
+				<SidebarItem on:click={() => console.log('Locale clicked in sidebar')}>
```
lkeegan, line 130:
```suggestion
				<SidebarItem>
```

lkeegan (approved):
lgtm! a couple of minor comments

frontend/src/lib/components/UserLandingPage.svelte
jmsssc: I agree with all of these and have made these changes. I'll merge once CI checks pass.

PR #239, opened 2025-03-11
Update DEVELOPMENT.md

jmsssc:
(Remove unnecessary workaround info)

@lkeegan I also wanted to check if it make sense to change `pip install .` to `pip install -e ` in the python install docs?

Only because with the former, the mondey_backend module runs with whatever the code was when first installing, with -e, it will change for the backend when users run 'mondey-backend' (Though I have the feeling I'm missing an easier way to do that)

But I haven't done as much Python recently so I'm not totally sure that's sensible. It just got editing backend python code to work for me just now.

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -84,14 +84,3 @@ pnpm run dev
 ```
 
 The website is then served at http://localhost:5173/.
-
-3.2: If you encounter CORS/login/API connection from frontend issues, do the below: Implement this CORS workaround so your frontend and backend requests connect after login (if you see HTTP 401 or "Unauthorised", it could be CORS)
-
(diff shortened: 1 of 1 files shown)


lkeegan:
> (Remove unnecessary workaround info)
> 
> @lkeegan I also wanted to check if it make sense to change `pip install .` to `pip install -e ` in the python install docs?
> 
> Only because with the former, the mondey_backend module runs with whatever the code was when first installing, with -e, it will change for the backend when users run 'mondey-backend' (Though I have the feeling I'm missing an easier way to do that)
> 
> But I haven't done as much Python recently so I'm not totally sure that's sensible. It just got editing backend python code to work for me just now.

Yes absolutely, an editable install makes sense.

lkeegan (approved):
thanks lgtm!

PR #247, opened 2025-03-18
186 try without login

jmsssc:
Hi @lkeegan and @MaHaWo so basically this branch is fully ready for a PR but I'm just not sure about these two questions, if either of you two have thoughts to share so I can quickly change them if changes are needed. But this is otherwise ready to review.

>     * **Should the anonymous test users have "isResearcher" and "isAdmin" set to true? Currently they only see personal data/kids pages** (just a json change)
> 
>     * **When the test user registers a permanent account, they technically stay logged in until they open the email, so they could mistakenly add data to the test account after registering unless we log them out**. Should we to be on the safe side log them out of the test user session, immediately after they sign up for a normal account, so they don't start adding permanent data?


jmsssc:
@lkeegan Thank you - I have just made that change and now this is ready for review. I had left something from protyping in there which has also been removed.

frontend/src/lib/translations.ts
```diff
@@ -18,45 +18,43 @@ export const translationIds = {
...
-		help: "Förderhilfen",
+		help: "Fördern",
 		info: "Information",
 		next: "Weiter",
 		prev: "Zurück",
 		moreInfoOnLegend: "Mehr Info",
 		autoNext: "Automatisch weiter",
 		groupOverviewLabel: "Übersicht Meilensteingruppen",
 		alertMessageRetrieving: "Fehler beim Abrufen der Meilensteine",
+		alertMessageNoRelevantMilestones: "Sie mussen zuerst Meilensteine zusetzen, die für das Alter des Kindes +/- 6 Monaten relevant gelten. Entfern des Kindes und nachdem Sie die Milesteine zusetzt haben, mach mal das hier wieder.",
```
MaHaWo, line 29: I think the german translation is a little bit off: It conveys to the user that they should add milestones that are relevant for the child, but they typically cannot do this because only admins can do that. 

I would suggest therefore: "There are currently no milestones available that would be relevant for the current child" - in german: 

```ts 
alertMessageNoRelevantMilestones: "Im Moment sind keine Meilensteine vorhanden, die für dieses Kind relevant wären", 
```
If you want to keep parts of the current translation, I would like to offer some improvements: 
- "Sie müssen... ". This is a more formal form of adress "Sie", but at the end you use 'Mach mal das hier wieder' -> this is an informal form of address, seen by adding the person explicitly: 'Du, mach mal das hier wieder'. These should not be mixed. 
- 'zusetzen' - this mostly means 'to clog up'. 'Add' would be 'hinzufügen'. 
- 'relevant gelten' - 'relevant' works like in english: 'relevant sind' (are relevant) instead of 'relevant gelten'. 
- 'Entfern des Kindes ... ' - switches to informal form of adress again, and should probably mean something like this (?) 'Leave this page and retry after milestones have been added'/'Please try again after milestones have been added' --> "Bitter versuchen sie es erneut, nachdem Meilensteine hinzugefügt wurden". 'Entfern ...' is also an informal imperative, but the next part should then be 'das Kind' (the child), not 'des Kindes' ('the child's).

mondey_backend/src/mondey_backend/main.py
```diff
@@ -21,21 +23,23 @@
...
 @asynccontextmanager
 async def lifespan(app: FastAPI):
     create_mondey_db_and_tables()
     await create_user_db_and_tables()
+    await scheduled_update_stats()
     scheduler = AsyncIOScheduler()
     scheduler.add_job(
-        scheduled_update_stats,
-        CronTrigger.from_crontab(app_settings.STATS_CRONTAB),
+        lambda: asyncio.run(scheduled_update_stats()), # Review appreciated of this approach
```
MaHaWo, line 41: I'm not the most familiar with asyncio, but from what I understand about SQLite, it should be ok. @lkeegan you have more insight here?

mondey_backend/tests/utils/test_auth_utils.py
```diff
@@ -0,0 +1,24 @@
+from mondey_backend.users import is_test_account_user
+
+
+# It would probably be better to use the User class, and add the email field to it, than to use a mock user
```
MaHaWo, line 4: this could perhaps become a pytest fixture?

```python 
@pytest.fixture
def test_user(): 
    return UserRead(
        id=x,
        email="123tester@testaccount.com",
        is_active=True,
        is_superuser=False,
        is_researcher=False,
        full_data_access=False,
        research_group_id=123451,
        is_verified=True,
    )
 ``` 
 
 and then go (careful, untested): 
  
 ```python
 def test_is_test_account(active_user, test_user):
       assert is_test_account_user(test_user)
       assert is_test_account_user(active_user) is False 
       ... 
```

mondey_backend/tests/utils/test_auth_utils.py
```diff
@@ -0,0 +1,24 @@
...
+def test_is_test_account():
+    tester_email = '123tester@testaccount.com'
+    nontester_email = 'heidelberguser@uni-heidelberg.de'
+    nontester_email_2 = 'Contester@gmail.com'
+
+    assert is_test_account_user(MockUser(email=tester_email))
+    assert False == is_test_account_user(MockUser(email=nontester_email))
+    assert False == is_test_account_user(MockUser(email=nontester_email_2))
+
+    try:
```
MaHaWo, line 20: there is also `with pytest.raises(...) that might be helpful here: 

```python 
with pytest.raises(TypeError, message = "... "): 
      do_stuff_here_that_raises_TypeError 
```
again, untested code...

mondey_backend/tests/routers/test_auth.py
```diff
@@ -129,3 +154,23 @@ def test_user_forgot_password_invalid_email(
     response = user_client.post("/auth/forgot-password", json={"email": email})
     assert "@" in response.json()["detail"][0]["msg"]
     assert response.json()["detail"][0]["type"] == "value_error"
+
+
+
+# It would probably be better to use the User class, and add the email field to it, than to use a mock user
+class MockUser:
```
MaHaWo, line 161: see comment on the other use of `MockUser`.

mondey_backend/src/mondey_backend/statistics.py
```diff
@@ -182,10 +188,20 @@ def _get_statistics_by_age(
 
     return count, avg, stddev
 
+async def get_test_account_user_ids(user_session: UserAsyncSessionDep) -> List[int]:
+    """
+        Lists user IDs that are identified as test accounts
+    """
+    users = await user_session.execute(select(User))
```
MaHaWo, line 195: doesn't this work? 
```python 
session.exec(select(User.id)).all() 
```
it's a little more idiomatic for getting lists of stuff from the database imho...

MaHaWo (approved):
good from my side, a few questions/suggestions here and there but nothing big.

MaHaWo:
I noticed, however, that I can´t build it anymore on my machine using the docker compose system... need to have a closer look tomorrow.

frontend/src/lib/components/UserLogin.svelte
```diff
@@ -28,6 +28,8 @@ async function refresh(): Promise<string> {
 	return "success";
 }
 
+export const refreshLogin = refresh()
```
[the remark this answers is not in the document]
lkeegan: @MaHaWo I think this is a "real" export in the normal javascript sense of "make this publicly visible to code that imports this file" rather than svelte's "magic" export meaning "make this reactive"

jmsssc:
> @jmsssc could you rebase on / merge the main branch? apart from the conflicts it looks like some commits here are already in main and unrelated to this PR? thanks!

Sure. I'll implement the suggestions that you and @MaHaWo have made (Thank you by the way for both the technical corrections and the German language explanations, I'll make some notes for this evening on that) and then merge main into this one so we can either rebase it onto main or merge it into main to conclude the pull request. I especially appreciate the backend comments on the DB (since my python SQL experience is mostly through ORM libraries/MongoDB) so it's great to know when there is a better approach.

frontend/src/lib/components/FrontPageAbout.svelte
```diff
@@ -1,7 +1,61 @@
...
+import {authCookieLogin, registerRegister} from "$lib/client/services.gen";
 import { i18n } from "$lib/i18n.svelte";
-import { Card } from "flowbite-svelte";
+import {Button, Card} from "flowbite-svelte";
+import type {AuthCookieLoginData, RegisterRegisterData, UserCreate, ValidationError} from "$lib/client/index.js";
+import {goto} from "$app/navigation";
+import {refresh} from "$lib/utils/Login";
+
+let showAlert = $state(false);
+let alertMessage = $state(
```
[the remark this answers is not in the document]
jmsssc: Agreed - for now, I will use a temporary alert as we do want an error if something goes wrong. It will be centralized in another branch.

Issue #243, opened 2025-03-17
Add functionality to mark questions for children and parents as mandatory


MaHaWo:
- add required field to the question definition in the admin interface
- add field in the database entry for that
- alternatively allow for them to be optional
- add error check to frontend to make sure the mandatory ones are filled.

PR #253, opened 2025-03-20
243 required option for questions

jmsssc:
Implemented, tested with Children questions. Need to test with parent questions.

Linked to issue #243


frontend/src/lib/components/ChildrenRegistration.svelte
```diff
@@ -450,7 +450,7 @@ async function submitData(): Promise<void> {
 								.additional_answer}
 							label={element?.text?.[i18n.locale].question}
 							textTrigger={element.additional_option}
-							required={element.component === 'fileupload' ? false : true}
+							required={(element.component === 'fileupload' || element.required === false) ? false : true}
```
lkeegan, line 453: it looks like the fileupload component created above already has `.required=false`, so I think this can now be simplified (I guess this logic predates a required field?):
```suggestion
							required={element.required}
```

lkeegan (approved):
one minor comment, otherwise lgtm!

frontend/src/lib/components/ChildrenRegistration.svelte
jmsssc: Agreed, it makes more sense this way. The actions of interacting with the interface and what it does will align.

Issue #249, opened 2025-03-18
confirm before deleting a child


lkeegan:
**Is your feature request related to a problem? Please describe.**
Currently clicking on delete for a child deletes them with no confirmation.

**Describe the solution you'd like**
We should as a minimum ask for confirmation that the user wants to delete the child, perhaps even with one of those annoying "type the name of the child to confirm" dialogs, or a second "are you sure - you'll lose all your data for this child!" confirmation to try to avoid this happening by accident, as currently the user loses all their data with this action.

We have an existing confirmation dialog in the admin interface.

lkeegan:
resolved by #254

PR #254, opened 2025-03-20
Add confirmation box to deleting a child

jmsssc:
This is ready for review, it uses the delete modal from the admin interface. Since the phrasing was open to use here, I kept it as is. Works really well that way!

Related to #249


lkeegan (approved):
lgtm! one comment - since this DeleteModal is now used beyond the admin part of the website maybe it should be moved up to the top-level components directory?

jmsssc:
> lgtm! one comment - since this DeleteModal is now used beyond the admin part of the website maybe it should be moved up to the top-level components directory?

Sure, this makes sense to me too. I've done that and will merge once done.

PR #259, opened 2025-03-21
Playwright with backend

jmsssc:
Pull request for CI-testing purposes...

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -113,7 +113,7 @@ jobs:
       # Run the logged-in UI tests
       - name: Run logged-in UI tests
         working-directory: ./frontend
-        run: pnpm test:ui
+        run: pnpm test:ui || exit 1
 
       - uses: actions/upload-artifact@v4
         if: ${{ !cancelled() }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,32 +34,6 @@ jobs:
           token: ${{ secrets.CODECOV_TOKEN }}
           fail_ci_if_error: true
           verbose: true
-  frontend-ui-tests:
-    runs-on: ubuntu-latest
-    defaults:
-      run:
-        working-directory: ./frontend
-    timeout-minutes: 30
-    steps:
-      - uses: actions/checkout@v4
-      - uses: pnpm/action-setup@v4
-        with:
-          version: 9
-      - uses: actions/setup-node@v4
-        with:
-          node-version: 22
-          cache: 'pnpm'
-          cache-dependency-path: frontend/pnpm-lock.yaml
-      - run: pnpm install
-      - run: pnpm build
-      - run: pnpm exec playwright install --with-deps
-      - run: pnpm test:ui
-      - uses: actions/upload-artifact@v4
-        if: ${{ !cancelled() }}
-        with:
-          name: playwright-report
-          path: frontend/playwright-report/
-          retention-days: 7
   backend-unit-tests:
     runs-on: ubuntu-latest
     defaults:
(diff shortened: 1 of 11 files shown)


.github/workflows/ci.yml
```diff
@@ -34,32 +34,6 @@ jobs:
...
-      - run: pnpm install
-      - run: pnpm build
-      - run: pnpm exec playwright install --with-deps
-      - run: pnpm test:ui
-      - uses: actions/upload-artifact@v4
-        if: ${{ !cancelled() }}
-        with:
-          name: playwright-report
-          path: frontend/playwright-report/
-          retention-days: 7
```
lkeegan, line 62: can we keep this bit? (it uploads the playwright report at the of the CI job)

.github/workflows/ci.yml
```diff
@@ -80,3 +54,63 @@ jobs:
...
+            echo "Waiting for backend to be ready... ($timeout seconds left)"
+            timeout=$((timeout-1))
+            sleep 1
+          done
+          echo "Backend is ready!"
+
+      # Run the logged-in UI tests
+      - name: Run logged-in UI tests
+        working-directory: ./frontend
+        run: pnpm test:ui
```
lkeegan, line 116: does this talk to the backend that you started? I would have thought we'd need to do something similar to the /api proxy you added here where playwright starts the frontend dev server:
https://github.com/ssciwr/mondey/blob/74cfef63adabbaa9d1cd474e72ea6306b49a3564/frontend/playwright.config.ts#L38-L41

lkeegan: also maybe we can run the built site instead of the dev server here, since we already built in earlier in the job?

lkeegan (commented):
thanks, this is a good start!

could we add a simple test that involves the backend to confirm that the frontend is able to talk to it? e.g. login with some username should display "incorrect username/password" message.

.github/workflows/ci.yml
jmsssc: > does this talk to the backend that you started? I would have thought we'd need to do something similar to the /api proxy you added here where playwright starts the frontend dev server:
> 
> https://github.com/ssciwr/mondey/blob/74cfef63adabbaa9d1cd474e72ea6306b49a3564/frontend/playwright.config.ts#L38-L41

Yes I believe so, because the Playwright logs for the tests appear and the anonymous user login spec is ran, which creates a guest account and checks that the svelte user store has its data. (as backend will be running from L95) - it won't need a proxy because the config that proxies :5173/api to :8000 will still work here, but the CI just checks at :8000 because the "proxy" on the frontend doesn't exist yet till we start the frontend (but we want to check backend is working before we try to start the frontend). The "&" on L95  should keep both commands alive continuously, keeping mondey-backend in the background, I believe.

I will create a false-positive test with the wrong user/password though as you mentioned that sounds like the best way to confirm what's working here.

lkeegan: Sounds good - is that test maybe not pushed to this branch? I do see one here but it's commented out: https://github.com/ssciwr/mondey/blob/main/frontend/tests/anonymoussignup.spec.ts

jmsssc: > Sounds good - is that test maybe not pushed to this branch? I do see one here but it's commented out: https://github.com/ssciwr/mondey/blob/main/frontend/tests/anonymoussignup.spec.ts

Yeah that's it, I just identified this too and I'm sorting it out. I believe it has something to do with translations too.

jmsssc: > Sounds good - is that test maybe not pushed to this branch? I do see one here but it's commented out: https://github.com/ssciwr/mondey/blob/main/frontend/tests/anonymoussignup.spec.ts

The test worked in another branch which I had been testing, but I hadn't activated it here. I've deleted the other branch to avoid confusion. I can't be sure, but the number of successful tests increased on the other branch, and all passed, so I thought it worked. However, I'll see now if CI on this branch works, with a false negative test too, and only then submit it for review again

jmsssc:
Not working yet despite passing tests. Failed tests do not fail the CI.

frontend/tests/anonymoussignup.spec.ts
```diff
@@ -1,22 +1,14 @@
...
-    expect(isLoggedIn).toBeTruthy();
-});
+// Could export some kind of "async logInAnonymousTestUser"
 
- */
+test("/ : Anonymous users can log in", async ({ page }) => {
+	await page.goto("/", { waitUntil: "networkidle" });
+	const button = page.getByTestId("anonymousLogin");
+	await button.click();
+	await page.pause();
```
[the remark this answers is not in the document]
jmsssc: Good spot yes, no longer is the page.pause() needed, I have removed it

PR #278, opened 2025-03-26
Ensure all necessary texts exist when a language is added

### LINKED ISSUE #275: Adding a language breaks milestones
**Describe the bug**
If an admin adds a new language to the site, the existing milestones and milestone groups do not have texts for that language.
This causes 404s/errors, e.g. the milestone cannot be edited in the admin interface anymore.

**To Reproduce**
Add a new language in the admin interface, then try to edit an existing milestone.

**Expected behavior**
When a new language is added, we also need to create the corresponding `Text` objects for all existing milestones, milestones groups and questions, where `Text` is:

https://github.com/ssciwr/mondey/blob/11f2d85dbd3d1f1ad8b7327a488e5c0c3cf034df/mondey_backend/src/mondey_backend/routers/utils.py#L41

Similarly when a language is removed these should all be deleted via cascade / relationship properties in the database.

lkeegan:
- Create all necessary texts when a new language is added
  - each `MilestoneGroup` needs a MilestoneGroupText
  - each `Milestone` needs a `MilestoneText`
  - each `UserQuestion` needs a `UserQuestionText`
  - each `ChildQuestion` needs a `ChildQuestionText`
- Delete these texts and the corresponding i18n json file when a language is deleted
- resolves #275

### DIFF
--- a/frontend/tests/research.spec.ts
+++ b/frontend/tests/research.spec.ts
@@ -85,7 +85,7 @@ test("Research page: valid data", async ({ page }) => {
 
 test("Research page: no data", async ({ page }) => {
 	await mock_user_route(page);
-	await page.goto("/research");
+	await page.goto("/userLand/research");
 
 	// no plot or table are displayed if there is no data
 	const plot = page.getByTestId("researchPlotLines");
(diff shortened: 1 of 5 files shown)


jmsssc (approved):
lgtm, worked when I added and removed a language, though I haven't ran the tests which I see passed on CI.

PR #282, opened 2025-03-27
Add test database and static files for e2e tests

### LINKED ISSUE #236: Playwright setup needs a way to test authentication protected pages.
- we are only testing the parts of the webpage that are not login protected right now 
- playwright needs access to authenticated state. Because we have no persistent database in the tests, this is difficult.
- only solution without creating weird test-only backdoors or whatever I see right now is to implement #186 first and use its functionality to test the pages that normally are login protected.

lkeegan:
- add top-level e2e folder with backend database for running e2e tests
  - user.db with user/researcher/admin users
  - mondey.db with a single child as proof-of-concept
- refactor frontend tests
  - tests/ui
    - self-contained ui tests, don't need a backend to run
    - add CI job to run these
    - modify page fixture to include mock languages routes that returns the frontend de translations
  - tests/e2e
    - e2e tests, need a backend with e2e test db
    - renamed existing frontend ui test CI job to e2e
  - add pnpm `test:e2e` and `test:e2e:dev` commands to run the e2e tests
- backend config improvements
  - remove `ENABLE_CORS` option: no longer needed to run locally since we now proxy the backend and frontend to have the same address
  - add `SMTP_HOST` to docker-compose args
  - add `COOKIE_SECURE`	to docker-compose args default of true
  - hard-code `RELOAD` arg to false in docker-compose
  - add `MONDEY_HOST` to app settings and docker-compose
- make backend email SMTP_HOST optional
  - if not set then the email message is logged instead of sent
  - useful when running e2e tests or running locally
- extend e2e tests
  - login tests check that user, researcher, admin can log in and that the appropriate navbar links are displayed
  - for the user they also check that their child is displayed
- resolves #236


jmsssc (approved):
This all looks great to me, I love that we can now change either direct SQL data to have it prepared for each test, or we can use the UI to test the interface and frontend-backend interactions. Works really well!

I had an issue running it for a second before realising the backend needs to be ran from e2e for the SQL data to be present when running e2e:dev locally. I don't think a warning or running `mondey-backend` from e2e directory make sense, because a running backend could be the intentional e2e one, or accidentally still using the main backend.

PR #283, opened 2025-03-27
Minor frontend tidy-up

lkeegan:
- deprecated $page -> page
- deprecated on:click -> onclick
- use obj?.x instead of obj.x if obj may be null
- add missing autocomplete fields to password inputs
- remove duplicated code
- remove some logging


jmsssc (approved):
All looks good to me

PR #284, opened 2025-03-28
Mangle data into sql database

jmsssc:
When provided the right files, this code will load in all the sample data with the correct milestone data, milestones meta data, children, parent users for each child, freetext answers which are independent. Lastly, this contains the contextual additional select answers (which sometimes have "freetext" optional answers, which when present are saved as "additional_answer" on the select).

All questions are assumed required, some based on labelling are addressed to parents and so stored as UserQuestions using user_id to link to the child.


jmsssc:
Hi @lkeegan , not sure if we want to merge this or just use it, but this is ready to have a look at if you'd like to! 

To run it, place the CSV files with the names used in `import_data/utils.py` near the top of the file:
```
# All 3 files come from SoSci Spreadsheet export...
labels_path = "labels_encoded.csv"  # originally codebook_MONDEY_0-6_2025-03-20_16-16.xlsx (export it to CSV first)
data_path = "data.csv"  # originally data_MONDEY_0-6_2025-03-20_16-16.csv
milestones_metadata_path = "milestones_metadata_(variables).csv"  # originally variables_MONDEY_0-6_2025-03-20_16-16.csv
# The other "Values" CSV SoSci file is not used in this process.
```

The main script to run from `import_dara` dir is:
`import_all.py true`

This should save 208 milestones and 7966 answers.
These will appear in `import_data/db/mondey.db` (and the users in `...users.db`)

There is further information in the README in `import_data`.

lkeegan:
thanks! just trying to run this locally, do I also need questions_specified.csv?

mondey_backend/src/mondey_backend/import_data/import_children_with_assigned_milestone_data.py
```diff
@@ -0,0 +1,214 @@
...
+                    "Are you sure you want to add further milestone data, possibly duplicating answers?"
+                )
+
+            # Create a milestone answer session for this child
+            answer_session = MilestoneAnswerSession(
+                child_id=child_id,
+                user_id=parents_id,
+                expired=True,
+                included_in_statistics=False,
+                created_at=datetime(2025, 1, 1, 1, 0, 1),
```
lkeegan, line 172: I suggested making this date 2025/01/01 but I just realised this is a very bad idea! When calculating the stats we determine how old the child was when they answered the questions for an answersession based on their date of birth and when the answer session was created. So this date really needs to be the (approx) date when the questionnaire data was collected - is that stored somewhere in the data? (by the way this would also explain why I am seeing negative child ages in the research page data)

lkeegan: @jmsssc looks like there's a `STARTED` column with the date we need for this - otherwise this data all looks great from playing around with it locally!

lkeegan: I updated the mondey.db you sent me using these STARTED values with this notebook: 5c2feb6

jmsssc:
Importing the latest email batch is going well - the only part that's a bit tricky is that the import script assumed the original questions would be there e.g.

the various kinds of Eltern were their own questions:

image

What I've done is temporarily recreated those questions with the same IDs they had originally as a hardcoded script, then I rerun the answer merging code (which we know must work exactly as it did before). **However this does not feel like a good solution really, even though it allows adding batches of data, the data we have actually received is the complete set each time (inclusive within it the previous data). Re-importing fresh every time is a bit more reproducible and reliable, and testable, so I am leaning towards  implementing it that way.**

So basically now the additional data import script:

1. Recreate the merged questions in their simple form (FP01, FP02 etc). This is achieved by rerunning the `import_childrens_question_answers_data` function which originally was not planned to run during additional data import. It will ignore existing questions, just adding in again the basic ones.
2. Insert all the children and new parents, ignoring duplicates
3. Insert the new childrens answers, into the simple question forms we just temporarily made (FP01 etc)
4. As part of "postprocess_corrections", the same code we ran on the first set of children gets ran to merge them.

In principle we could also import all in one batch again from scratch without dealing with the updating issue.

jmsssc:
@lkeegan this works fully from the frontend through to the backend for the user, importing the data, updating them and erroring if the data has a bad format. It removes duplicates in the CSV before processing it. In REFACTORING.md there are some major improvements some of which might be worth doing despite that this is less every-day API code and more a once-every-now-and-then piece of code. I will continue to improve the quality and get the tests working with the async changes I had to make to connect the frontend form to it, which will make it more sustainable in the long run (the tests in principle work, once connected).

jmsssc:
I also plan to check:
- [ ] Are the STARTED values for MilestoneAnsweringSessions the values from the data? (as with the alter update)?

mondey_backend/src/mondey_backend/routers/research.py
```diff
@@ -46,4 +57,83 @@ async def get_research_names(session: SessionDep):
...
+        try:
+            contents = await file.read()
+            csv_data = pd.read_csv(
+                io.BytesIO(contents),
+                sep="\t",
+                encoding="utf-16",
+                encoding_errors="replace",
+            )
+
+            print("Read file.")
```
lkeegan: print -> logging

mondey_backend/src/mondey_backend/routers/research.py
```diff
@@ -46,4 +57,83 @@ async def get_research_names(session: SessionDep):
...
+                    data_path=csv_file,
+                    labelling_path="src/mondey_backend/import_data/labels_encoded.csv",
+                    questions_configuration_path="src/mondey_backend/import_data/questions_specified.csv",
+                )
+                # if it errors, it will go through to the throw 400 error block.
+                print("Finished adding additional data")
+
+                return {
+                    "status": "success",
+                    "message": "CSV data successfully imported",
```
lkeegan: it would be nice to tell the user how many rows/answer sessions were imported

lkeegan (commented):
UI looks good, made a couple minor suggestions, although I didn't get it to import successfully locally so far..

mondey_backend/src/mondey_backend/import_data/manager/data_manager.py
```diff
@@ -0,0 +1,262 @@
...
+        # Database paths
+        self.current_db_path = current_db_path or (
+            script_dir / "src/mondey_backend/import_data/current_db/current_mondey.db"
+        )
+        self.users_db_path = users_db_path or (
+            script_dir / "src/mondey_backend/import_data/current_db/current_users.db"
+        )
+
+        # Database connections
+        # When we want this to work directly on the live DB, we can change these paths to be the same as the normal ones.
```
lkeegan, line 122: when using the live db the endpoint can just pass the `session` and `user_session` deps directly - then we don't need to create the engines or the sessions here

lkeegan (commented):
thanks - this looks good!

One suggestion (which I think you already mentioned) would be to first construct all the objects for a single row (i.e. User, Child, AnswerSession, Child/User/MilestoneAnswers, etc) - then if it all worked add them to the database, then go to the next row. This would avoid the danger of half-importing data if the import e.g. first creates all the children & milestone answers, then runs into an error importing the answers.

Also a general question: do we ever want to do a full import again, or use this as a CLI?
My person answer would be no: we only need to be able to add additional data using the web interface. If this is the case then quite a lot of the code here could be removed from this PR (we could still keep this branch) - reducing the maintenance / testing burden. What do you think?

jmsssc:
> thanks - this looks good!
> 
> One suggestion (which I think you already mentioned) would be to first construct all the objects for a single row (i.e. User, Child, AnswerSession, Child/User/MilestoneAnswers, etc) - then if it all worked add them to the database, then go to the next row. This would avoid the danger of half-importing data if the import e.g. first creates all the children & milestone answers, then runs into an error importing the answers.
> 
> Also a general question: do we ever want to do a full import again, or use this as a CLI? My person answer would be no: we only need to be able to add additional data using the web interface. If this is the case then quite a lot of the code here could be removed from this PR (we could still keep this branch) - reducing the maintenance / testing burden. What do you think?

I think both these two core changes are doable and make sense. I'll do these and test again that import works, and then switch to live DB connection + other small changes

jmsssc:
Now reviewing code quality and tests.

mondey_backend/src/mondey_backend/import_data/manager/import_manager.py
```diff
@@ -0,0 +1,1297 @@
...
+
+    async def import_children_with_milestone_data(
+        self, session: Session, data_df
+    ) -> None:
+        """Import children with milestone data."""
+        logger.info("Importing children with milestone data")
+
+        # Get all milestones
+        milestone_query = select(Milestone)
+        milestones = session.execute(milestone_query).scalars().all()
```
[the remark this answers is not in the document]
lkeegan: Yup the fact that we use the sqlmodel for mondey.db but have to use sqlalchemy for user.db is horrible, I'll definitely try to avoid that happening in future projects!..

jmsssc:
@lkeegan this is all working again with much cleaner tests and a much thinner import_manager and data_manager as a result of being additional-data only focused with fixtures/stub data for the import tests.

I am fixing the other tests and addressing the 2 issues[1] but it would be great to see if you have any review thoughts on the code changes & import_manager/data_manager specifically. The remove_duplicates test is not complete yet, but the import manager test file with its 6 tests is complete.

[1]
[3.2% Duplication on New Code](https://sonarcloud.io/component_measures?id=ssciwr_mondey&pullRequest=284&metric=new_duplicated_lines_density&view=list) (required ≤ 3%)
[C Reliability Rating on New Code](https://sonarcloud.io/dashboard?id=ssciwr_mondey&pullRequest=284) (required ≥ A)

mondey_backend/src/mondey_backend/import_data/manager/import_manager.py
```diff
@@ -0,0 +1,952 @@
...
+
+        # Get sessions
+        import_session, _ = self.data_manager.get_import_session()
+
+        try:
+            # Import additional data
+            await self.import_additional_data(import_session)
+
+            logger.info("Additional data import completed successfully")
+
```
lkeegan, line 934: My understanding is that rollback only rolls back uncommited changes - once you commit there is no going back.

So I think we would need to not do *any* commits in the code here, then do a single commit after import_additional_data ran succcessfully, and otherwise catch any exception and rollback (as done currently)

https://docs.sqlalchemy.org/en/20/orm/session_basics.html#framing-out-a-begin-commit-rollback-block

lkeegan: Maybe replacing all the `.commit` calls with `.flush` would work?

mondey_backend/src/mondey_backend/import_data/manager/import_manager.py
```diff
@@ -0,0 +1,952 @@
...
+        # Then import answers
+        self.import_answers(session, additional_data_df)
+
+        logger.info("Additional data imported successfully")
+
+    # -------------------------------------------------------------------------
+    # Import Method
+    # -------------------------------------------------------------------------
+
+    async def run_additional_data_import(self) -> None:
```
lkeegan, line 922: it would be nice for this function to return some information about what was done, e.g. a string description, or at least an int with the number of imported answer sessions (which can then be passed on to the router + frontend)

mondey_backend/src/mondey_backend/import_data/manager/import_manager.py
lkeegan: Also a related point: I would pass in the session and user_session as arguments everywhere and then call this code from the router passing in the session and user_session. If you instead create a session within this code, then once it goes out of scope `.commit` is implicitly called, so I think it's safer to avoid doing this if we want to be able to rollback all the changes. (Also we should rollback the user_session as well)

lkeegan (commented):
Thanks, the data/import manager changes looks good! I made some suggestions, mostly on the rollback / sessions

mondey_backend/src/mondey_backend/import_data/manager/data_manager.py
```diff
@@ -0,0 +1,236 @@
...
+        self.load_labels_df(force_reload=True)
+
+    # -------------------------------------------------------------------------
+    # Database Session Methods
+    # -------------------------------------------------------------------------
+
+    def get_import_session(self) -> tuple[Session, Engine]:
+        with Session(mondey_engine) as session:
+            return session, mondey_engine
+        # todo: change to just use the normal session in code that currently calls this?
```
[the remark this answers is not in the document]
jmsssc: Sure. I'll update the tests to pass the test session (from conftest) via the ImportManager as an arg rather than overwriting the method in data_manager here as it does in test_import_additional.py `manager.data_manager.get_import_session = lambda: (session, engine)`

mondey_backend/src/mondey_backend/import_data/manager/import_manager.py
jmsssc: Yes sorry, that is the next thing I had started to do before focusing to make sure the 6 tests would test what we want and pass locally with the user providing labels/data CSVs via the Data page.

I will add that to this PR.

mondey_backend/src/mondey_backend/import_data/manager/import_manager.py
```diff
@@ -0,0 +1,952 @@
...
+
+        user_create = UserCreate(
+            email=email,
+            password="$$$$testUser$$$$432hjdfioj3409lk",
+            is_researcher=False,
+            full_data_access=False,
+            research_group_id=0,
+        )
+
+        user = User(
```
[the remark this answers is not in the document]
jmsssc: Sure

mondey_backend/src/mondey_backend/import_data/postprocessing_corrections/run_postprocess_corrections.py
```diff
@@ -0,0 +1,33 @@
...
+from mondey_backend.import_data.postprocessing_corrections.update_started_value_for_answersessions import (
+    update_started_value_for_answersessions,
+)
+from mondey_backend.import_data.utils import data_path
+
+
+# cleanup_question is really functioning as "are you running this for the first time", this function needs
+# to be refactored into the two main versions of it really, and the question deletion/merging stuff should only
+# happen when you run it for the first time.
+def run_postprocessing_corrections(
```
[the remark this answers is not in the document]
jmsssc: This has been done ✓

mondey_backend/src/mondey_backend/import_data/manager/import_manager.py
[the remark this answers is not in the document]
jmsssc: I have made this change

jmsssc:
@lkeegan this is now reviewable and the 6 tests should work:

Beware they take quite some time to run, because each one imports everything. To run these you need the labels_encoded.csv in your `import_data` directory(should be there from when you last reviewed this and I sent it to you) and `additional_data_full_backup.csv` in that same directory (this is just a copy of the additional_data.csv - it should have 427 rows. You may need to redownload the additional_data.csv I sent you last time if your one has been modified in place by the deduplication code (to either 106 or 0 rows))

The remove duplicates test is also working and ready.

```
$  python3 -m pytest -x tests/import_data/test_import_additional.py -s --log-cli-level=DEBUG
```
============================================================================================== 
6 passed in 286.64s (0:04:46) ==============================================================================================

lkeegan (changes requested):
thanks, this looks good! I push a couple of minor fixes to get the frontend to work locally, then it seemed to upload the dataset successfully as far as I can tell.
Made some more suggestions (sorry, nearly done I promise!) and there are a couple of suggestions still open from the last review.

Re. commit vs flush, it still seems like we can end up with a partial import currently, as e.g. the child questions are committed before the milestone answers (or vice versa, but the point is there are multiple commits). It would be nice if the import process was atomic - could we just replace all the commits with flushes and have a single commit or rollback to both sessions at the end?

jmsssc:
I will update this and sort this out. It was working well on Friday but I'd like to take a fresh look today first, go over your suggestions again to double check those, and I am doing other tasks first. I will @ you Liam whenever it's ready for review.

jmsssc:
Resolving conflicts now, then testing the UI upload - then this should be ready for review.

jmsssc:
image

frontend/src/lib/components/Admin/AnswerSessionData.svelte
```diff
@@ -70,6 +83,93 @@ async function answerSessionAnalysisModalCallback() {
...
+			stats_out_of_date = true;
+			await refreshMilestoneAnswerSessions();
+		}
+	} catch (e) {
+		console.error(e);
+		importResult = {
+			status: "error",
+			message: e.message || i18n.tr.admin.importFailed,
+			error: true,
+		};
```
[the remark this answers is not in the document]
jmsssc: image

works

jmsssc:
@lkeegan this is ready for review.

> ... Re. commit vs flush, it still seems like we can end up with a partial import currently, as e.g. the child questions are committed before the milestone answers (or vice versa, but the point is there are multiple commits). It would be nice if the import process was atomic - could we just replace all the commits with flushes and have a single commit or rollback to both sessions at the end?

I have implemented this too:
image

mondey_backend/src/mondey_backend/import_data/manager/import_manager.py
```diff
@@ -0,0 +1,949 @@
...
+            children_imported = await self.import_additional_data()
+
+            logger.info(
+                f"Additional data import completed successfully. {children_imported} children imported."
+            )
+            return children_imported
+
+        except Exception as e:
+            logger.error(f"Error during additional data import: {e}")
+            self.data_manager.session.rollback()
```
lkeegan, line 948: I think there should also be a user_session rollback here?

lkeegan (approved):
Thanks, this looks great and works for me locally!

One new comment and there are a still a few (minor) open ones from previous review rounds, otherwise this looks good to go.

jmsssc:
> Thanks, this looks great and works for me locally!
> 
> One new comment and there are a still a few (minor) open ones from previous review rounds, otherwise this looks good to go.

Sorry. I now have a better way to see all your suggestions, as TLDR, I had only been seeing some of them the way I was going through them.

jmsssc:
I know you approved this but would you like a last look at https://github.com/ssciwr/mondey/pull/284/commits/a4aa5ba310440b60003f807bcdee95455d908b7d ?

jmsssc:
I noticed one remaining issue - there seem to be missing answers. I am going to investigate and then add a test case.

PR #285, opened 2025-03-31
Standardize tailwind css buttons

### LINKED ISSUE #235: Clean up tailwindcss classes
The current repetitive tailwindcss usage we have in the frontend is a nightmare to maintain, and needs to be cleaned up 
- extract and unify common patterns
- make use of @apply 
- for more, see [here](https://v3.tailwindcss.com/docs/reusing-styles#extracting-classes-with-apply)

jmsssc:
Now the tailwind theme uses the blue/green teal color. Interactive parts of the site use this color, including <Buttons>, <buttons>, some form element outlines, and <a links in text form (e.g. "Forgot password" on the login page). Future flowbite svelte components, tables, accordions etc, should all work with this theming.

Additionally, milestone buttons remain with the "Blue" color.

I prioritized using tailwind theming for future flowbite component use compatibility, but some exceptions remain hardcoded in app.css, mostly for icons, a media query, and app basics. I also felt setting up SCSS might not be worth it. The tradeoff with that is that theming works great, but we are not using variables as much as ideal (For example, some hardcoded colors in the tailwind theme config are repeated).

This will close #235

### DIFF
--- a/frontend/tailwind.config.ts
+++ b/frontend/tailwind.config.ts
@@ -9,6 +9,7 @@ export default {
 		"./node_modules/flowbite-svelte/**/*.{html,js,svelte,ts}",
 	],
 	darkMode: "selector",
+	// ensure svelte compiler doesn't optimize these away if it doesn't realise we are using them
 	safelist: [
 		"bg-milestone-answer-0",
 		"bg-milestone-answer-1",
--- a/frontend/tailwind.config.ts
+++ b/frontend/tailwind.config.ts
@@ -21,7 +21,7 @@ export default {
 				// From your original additional-color in CSS
 				primary: {
 					50: "#E2F1F0",
-					100: "#D5EAE9",
+					100: "rgb(50,50,50)",
 					200: "#C7E3E2",
 					300: "#B6DBDA",
 					400: "#A5D2D1",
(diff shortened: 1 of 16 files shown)


frontend/tailwind.config.ts
```diff
@@ -9,7 +9,6 @@ export default {
 		"./node_modules/flowbite-svelte/**/*.{html,js,svelte,ts}",
 	],
 	darkMode: "selector",
-	// ensure svelte compiler doesn't optimize these away if it doesn't realise we are using them
```
lkeegan, line 12: could we keep this comment? (otherwise this safelist looks pointless and I would probably delete it at some point in the future :-))

lkeegan (approved):
thanks - lgtm!

one question: should we add mouseover text to the edit and delete buttons (like the "anmeldung" button has on the front page), or do you think they are self-explanatory?

jmsssc:
> thanks - lgtm!
> 
> one question: should we add mouseover text to the edit and delete buttons (like the "anmeldung" button has on the front page), or do you think they are self-explanatory?

Yeah, I think that sounds useful, especially since we have them abstracted into one place (minus on the Kinder single view).

PR #289, opened 2025-04-01
Improve and document email deployment setup

### LINKED ISSUE #288: Emails are not sent outside of university network
This is due to uni firewall rules

lkeegan:
- backend
  - add username/password for smtp server to settings
    - if provided, use them when authenticating with the SMTP server
- docker-compose.yaml
  - add new settings variables
  - set postfix image to autogenerate DKIM keys
- deployment docs
  - how to use another SMTP server to send emails
  - how to use a relay host if using the docker postfix image to send emails
  - how to configure SPF/DKIM/DMARC DNS settings
- resolves #288


jmsssc (approved):
All makes sense to me, reading through the docs here.

PR #290, opened 2025-04-02
Minor changes to support imported survey data

lkeegan:
- backend
  - add `name` to Milestone database model
- admin page
  - add name to editable fields for a Milestone in EditMilestoneModal
- research page
  - ignore any data with age outside of range 1-72 when plotting
  - only do groupby operation on selected milestone for now to make UI more reponsive
- add `/admin/update-stats` endpoint
  - allows an admin to trigger the stats update manually in addition to the regular cron job update
  - can either do incremental update with only new data, or full re-calculation using all data
  - only API for now, i.e. no button to do this in the admin web interface
- see #284 for the code used to import the survey data

### DIFF
--- a/frontend/src/lib/components/Admin/EditMilestoneModal.svelte
+++ b/frontend/src/lib/components/Admin/EditMilestoneModal.svelte
@@ -19,6 +19,7 @@ import { i18n } from "$lib/i18n.svelte";
 import {
 	Button,
 	ButtonGroup,
+	Input,
 	InputAddon,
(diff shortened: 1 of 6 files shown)


jmsssc (approved):
All looks good to me, the only thing I thought about differently was making "name" an optional field. Because, it can be used to indicate that the data has came from some kind of survey/import rather than being manually added (at least some of). However having it be optional introduces complexity so this does all look great now to me.

PR #298, opened 2025-04-03
Keep UI active during research processing and use loader

jmsssc:


jmsssc:
This even appears to me to be slightly faster (I am not sure why, probably because it runs on the Webworker maybe on a different thread, whereas on Svelte, everything (e.g. checking whether $effect or $derived should react) may happen on one thread(if like React?)). But the long operations were the mean, std etc, those shouldn't be too different. It does seem to respond in < 1.5 seconds rather than the previous 2.5-3 seconds for the real data). But as data loading, until your changes, takes longer, it's not major.

lkeegan:
sounds good - is this ready for review?

jmsssc:
> sounds good - is this ready for review?

Yes. I was just looking at your PR.

Will your changes (since I saw the research page changed too, with probably faster calculations on the frontend) make this PR unnecessary? Or should I merge my PR into this PR so that the frontend can work without freezing during calculations?

lkeegan:
Let's first get your PR reviewed and merged, then I'll rebase mine

frontend/src/routes/userLand/research/dataWorker.ts
```diff
@@ -0,0 +1,184 @@
...
+			df_in,
+			milestoneId,
+			msg.selectedColumns,
+		);
+		console.log("W: Done processing.", columns);
+
+		// Send everything back to the main thread
+		const response: WorkerResponse = {
+			type: "dataProcessed",
+			dfOut: dfd.toJSON(df_out),
```
lkeegan, line 171: I think we can remove df_out here, it's not needed on the ui side

frontend/src/routes/userLand/research/+page.svelte
```diff
@@ -174,69 +194,95 @@ function generateFakeData() {
...
+	}
+	return Object.keys(json_data[0]);
+});
 </script>
 
 <div class="w-full grow">
-<!-- <TODO>: remove this once production site has real data-->
-<div class="flex flex-col m-2 bg-red-300">
-    Temporary FAKE data generation:
+    <!-- Fake data generation UI -->
```
lkeegan, line 217: I think we can remove all this fake data stuff now that we have real data in production

frontend/src/routes/userLand/research/+page.svelte
```diff
@@ -23,138 +24,157 @@ import {
...
-	selected_milestone_id = milestone_ids[0].value;
 }
 
 async function getData() {
-	const res = await getResearchData();
-	if (res.error || !res.data) {
-		console.error(res.error);
-		return;
+	console.log("Loading data from API");
+	try {
```
lkeegan, line 128: Could the serviceworker also get the data initially? Seems like we don't need it in the ui code, so instead of getting it here then sending it to the worker could the worker just get the data directly itself?

lkeegan (changes requested):
Thanks, looks good, a few suggestions/ comments

frontend/src/routes/userLand/research/dataWorker.ts
jmsssc: Sure

frontend/src/routes/userLand/research/+page.svelte
jmsssc: Will do

frontend/src/routes/userLand/research/+page.svelte
jmsssc: This is what I did for a while when I tried a +page.server.js approach using server actions (didn't do that because the imports basically didn't work and it seemed like an anti pattern to use serverActions without a cause when people might know web workers better). And I got stuck with some infinite re-renders but yeah now that is solved I will improve this.

I will change it and make it use await/then as you suggested before.

jmsssc: > This is what I did for a while when I tried a +page.server.js approach using server actions (didn't do that because the imports basically didn't work and it seemed like an anti pattern to use serverActions without a cause when people might know web workers better). And I got stuck with some infinite re-renders but yeah now that is solved I will improve this.
> 
> I will change it and make it use await/then as you suggested before.

Actually in this case I don't think using the await / then pattern makes sense, here is the thing:
- **onMount contains the worker set up and the receiving function (so would need to Promise.resolve this in the function we await**
- Currently another function deals with reacting to filter changes by filtering the data with the webworker, by emitting a worker event
- So basically, because we need onMount to set up the worker, but we can't really define the promise we want to await in onMount (and it needs to be there to get the response), with this pattern I think we should stick to updating state on getting the response from the web worker.

jmsssc: And since the API proxy does not work for webworker requests, at least using the autogenerated server.gen.ts function won't work, because it's missing the api proxy bit '/api':

image

So we could set up an API call to call it from the API (not sure if production backend will use the same-server proxy approach or a different server), but for now I will keep it on the frontend, however, if it makes the webworker slower transferring in the future, we can resolve it. I think the webworker might deal with it quite well though, on intuition. Because clientside data download + webworker processing is faster than what we had before.

lkeegan: ok then lets leave this as is for now (by the way the reason the client doesn't work might be that the webworker doesn't include layout.ts which sets the base url with the /api bit):

https://github.com/ssciwr/mondey/blob/084bd739556966d299e0b605b9fba57ace1fc400/frontend/src/routes/%2Blayout.ts#L4-L6

lkeegan (approved):

PR #299, opened 2025-04-03
293: Add tooltips to Edit and Delete buttons: Hovering over edit thou…

jmsssc:
…gh, the tooltip makes the delte button shift to the right.


lkeegan (approved):
lgtm!

PR #306, opened 2025-04-07
Don't display options in admin questions table as they can be very lo…

lkeegan:
…ng strings


jmsssc (approved):
Sure lgtm and a short solution to this issue

Issue #273, opened 2025-03-24
Deleted children answer sessions remain


lkeegan:
**Describe the bug**
Currently if a child is deleted, their answer sessions remain

We should

1. check that e.g. stats don't 404 / crash because of this -> add tests for this & fix if needed
2. decide if we want to cascade delete answer sessions (and child questions) when a child is deleted -> probably yes?

jmsssc:
@lkeegan Number 2 makes sense to me. Because it would make it more reliable in the long run (... to cascade and delete childanswers for a deleted child as well as child milestoneanswersessions).

We can additionally add a note on that when deleting children on the frontend.

I can only imagine when a child is deleted, someone has withdrawn consent for the data to be used, or the child is leaving the place, or was an outlier or so. Not being a domain expert, I cannot imagine when you would delete a child, but want the data on them to remain.

jmsssc:
**Secondly, I just removed some questions and then noticed that for statistics, this can cause similar key errors e.g.**

  File "Work/mondey/mondey_backend/src/mondey_backend/statistics.py", line 546, in get_child_answers
    child_answers[answer.child_id][questions[answer.question_id]] = answer.answer
                                   ~~~~~~~~~^^^^^^^^^^^^^^^^^^^^
KeyError: 21

(q 21 was deleted).

Should we delete all answers (user|child) referring to a question upon deleting the question?

lkeegan:
Yes I think for consistency this would make sense.

We just need to make it very clear in the UI that deleting a question in the admin interface implies any associated data will get deleted!

jmsssc:
Done in PR #309  (for issue #273)

> We just need to make it very clear in the UI that deleting a question in the admin interface implies any associated data will get deleted!

The user gets told how many individual answers will be deleted (e.g. for most questions 325 answers in the sample data, shown in a red box) and then they have to confirm they want to delete. They are asked to type "Löschen" for questions (or their languages equivalent translation), then they can confirm. It would be possible to require them to type the question but that can become quite long.

PR #309, opened 2025-04-07
273 - Cascade deletes for User & Child questions, milestones & milestone groups (with UI confirmation)

jmsssc:
@lkeegan the current code implements and tests deletion for all 4 object types in the backend.

I used a different approach for user and child questions (normal cascade) as to milestones/answers. Would be glad to hear your feedback on which approach is better while I complete the frontend confirmation and dry_run functionality to inform the user what they will be deleting.

Do you prefer one way (like using cascades totally) over the other way (for milestones/milestone answers: it manually executes delete queries on related milestones/answers). I was a bit more worried that the performance could be worsened with milestone groups/milestones having cascades to answers, as compared to just questions-answers which are more 1:1 (e.g. if you delete one of the five milestone groups, it might be quite a substantial query for the database?)

The cascade approach uses less code, and is pretty clear, but like you mentioned regarding language, I think it is a bit more risky on the performance side.

### DIFF
--- a/mondey_backend/src/mondey_backend/models/questions.py
+++ b/mondey_backend/src/mondey_backend/models/questions.py
@@ -4,6 +4,9 @@ from sqlalchemy.orm import Mapped
 from sqlmodel import Field
 from sqlmodel import SQLModel
 
+from .utils import back_populates
+
+# Different format to dict-relationship
 from .utils import dict_relationship
 from .utils import fixed_length_string_field
 
@@ -55,6 +58,9 @@ class UserQuestionText(QuestionTextBase, table=True):
 class UserQuestion(Question, table=True):
     id: int | None = Field(default=None, primary_key=True)
     text: Mapped[dict[str, UserQuestionText]] = dict_relationship(key="lang_id")
+    answers: Mapped[list[UserAnswer]] = back_populates(
+        "question", cascade="all, delete-orphan"
+    )
 
 
 class UserQuestionPublic(QuestionPublic):
@@ -80,6 +86,9 @@ class ChildQuestionText(QuestionTextBase, table=True):
 class ChildQuestion(Question, table=True):
     id: int | None = Field(default=None, primary_key=True)
     text: Mapped[dict[str, ChildQuestionText]] = dict_relationship(key="lang_id")
+    answers: Mapped[list[ChildAnswer]] = back_populates(
+        "question", cascade="all, delete-orphan"
+    )
 
 
 class ChildQuestionPublic(QuestionPublic):
@@ -105,6 +114,7 @@ class UserAnswer(AnswerBase, table=True):
     question_id: int = Field(
         default=None, primary_key=True, foreign_key="userquestion.id"
     )
+    question: UserQuestion = back_populates("answers")
 
 
 class UserAnswerPublic(AnswerPublicBase):
@@ -116,6 +126,7 @@ class ChildAnswer(AnswerBase, table=True):
     question_id: int = Field(
         default=None, primary_key=True, foreign_key="childquestion.id"
     )
+    question: ChildQuestion = back_populates("answers")
 
 
 class ChildAnswerPublic(AnswerPublicBase):
--- a/mondey_backend/src/mondey_backend/routers/admin_routers/milestones.py
+++ b/mondey_backend/src/mondey_backend/routers/admin_routers/milestones.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 from fastapi import APIRouter
 from fastapi import HTTPException
+from fastapi import Query
 from fastapi import UploadFile
 from sqlmodel import col
 from sqlmodel import select
@@ -23,6 +24,9 @@ from ...models.milestones import SubmittedMilestoneImagePublic
 from ...models.utils import ItemOrder
 from ...statistics import async_update_stats
 from ..utils import add
+from ..utils import count_milestone_answers_for_milestone
+from ..utils import delete_milestone_by_id
+from ..utils import delete_milestones_with_group_id
 from ..utils import get
 from ..utils import milestone_group_image_path
 from ..utils import milestone_image_path
@@ -69,11 +73,52 @@ def create_router() -> APIRouter:
         return db_milestone_group
 
     @router.delete("/milestone-groups/{milestone_group_id}")
-    def delete_milestone_group_admin(session: SessionDep, milestone_group_id: int):
+    def delete_milestone_group_admin(
+        session: SessionDep,
+        milestone_group_id: int,
+        dry_run: bool = Query(
+            True,
+            description="When true, shows what would be deleted without actually deleting",
+        ),
+    ):
         milestone_group = get(session, MilestoneGroup, milestone_group_id)
-        session.delete(milestone_group)
-        session.commit()
-        return {"ok": True}
+
+        if not dry_run:
+            affected_milestone_answers = 0
+            deleted_milestone_ids = delete_milestones_with_group_id(
+                session, milestone_group_id, dry_run=False
+            )
+            for milestone_id in deleted_milestone_ids:
+                affected_milestone_answers += count_milestone_answers_for_milestone(
+                    session, milestone_id
+                )
+
+            session.delete(milestone_group)
+            session.commit()
+            return {
+                "ok": True,
+                "deletion_executed": True,
+                "deleted_milestone_count": len(deleted_milestone_ids),
+                "deleted_answer_count": affected_milestone_answers,
+            }
+
+        affected_answers = 0
+        for milestone in milestone_group.milestones:
+            if milestone.id is None:
+                continue
+            affected_answers += count_milestone_answers_for_milestone(
+                session, milestone.id
+            )
+        return {
+            "ok": True,
+            "dry_run": True,
+            "would_delete": {
+                "milestone_ids": [
+                    milestone.id for milestone in milestone_group.milestones
+                ],
+                "affected_answers_count": affected_answers,
+            },
+        }
 
     @router.post("/milestone-groups/order/")
     def order_milestone_groups_admin(session: SessionDep, item_orders: list[ItemOrder]):
@@ -112,11 +157,32 @@ def create_router() -> APIRouter:
         return db_milestone
 
     @router.delete("/milestones/{milestone_id}")
-    def delete_milestone(session: SessionDep, milestone_id: int):
-        milestone = get(session, Milestone, milestone_id)
-        session.delete(milestone)
-        session.commit()
-        return {"ok": True}
+    def delete_milestone(
+        session: SessionDep,
+        milestone_id: int,
+        dry_run: bool = Query(
+            True,
+            description="When true, shows what would be deleted without actually deleting",
+        ),
+    ):
+        if not dry_run:
+            affected_milestone_answers = delete_milestone_by_id(session, milestone_id)
+
+            return {
+                "ok": True,
+                "deletion_executed": True,
+                "deleted_milestone_id": milestone_id,
+                "deleted_answer_count": affected_milestone_answers,
+            }
+        else:
+            affected_answers = count_milestone_answers_for_milestone(
+                session, milestone_id
+            )
+            return {
+                "ok": True,
+                "dry_run": True,
+                "would_delete": {"affected_answers_count": affected_answers},
+            }
 
     @router.post("/milestones/order/")
     def order_milestones_admin(session: SessionDep, item_orders: list[ItemOrder]):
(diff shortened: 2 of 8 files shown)


jmsssc:
> > @lkeegan the current code implements and tests deletion for all 4 object types in the backend.
> > I used a different approach for user and child questions (normal cascade) as to milestones/answers. Would be glad to hear your feedback on which approach is better while I complete the frontend confirmation and dry_run functionality to inform the user what they will be deleting.
> > Do you prefer one way (like using cascades totally) over the other way (for milestones/milestone answers: it manually executes delete queries on related milestones/answers). I was a bit more worried that the performance could be worsened with milestone groups/milestones having cascades to answers, as compared to just questions-answers which are more 1:1 (e.g. if you delete one of the five milestone groups, it might be quite a substantial query for the database?)
> > The cascade approach uses less code, and is pretty clear, but like you mentioned regarding language, I think it is a bit more risky on the performance side.
> 
> So I don't mind which is faster for deleting stuff, since that is a pretty rare occurence.
> 
> My concern is that adding relationships might make other unrelated things (that happen more often) slower - e.g. it looks like here if I get a `UserQuestionPublic` from the database I'll also now get a list with all the answers to that question, i.e. the same request now returns more data and is a bit slower? I'm not sure about this though, maybe the list is lazily fetched and it doesn't make any difference to the existing queries - I'll take a better look at this tomorrow.

This is the first time I've used a project with _Public for some returned calls, but it looks like it won't return answers by default to me. I could definitely be wrong though and that is absolutely something to be mindful about. I have just quickly debugged that by adding prints and deliberately failing the `test_get_user_question_admin_works` which returns the list of questions. That has UserQuestionPublic, but includes no answers, this isn't foolproof though:

image

The non-admin URI appears the same, answers aren't included:
image

Nor are answers included in the OpenAPI schema:
image

lkeegan:
> > > @lkeegan the current code implements and tests deletion for all 4 object types in the backend.
> > > I used a different approach for user and child questions (normal cascade) as to milestones/answers. Would be glad to hear your feedback on which approach is better while I complete the frontend confirmation and dry_run functionality to inform the user what they will be deleting.
> > > Do you prefer one way (like using cascades totally) over the other way (for milestones/milestone answers: it manually executes delete queries on related milestones/answers). I was a bit more worried that the performance could be worsened with milestone groups/milestones having cascades to answers, as compared to just questions-answers which are more 1:1 (e.g. if you delete one of the five milestone groups, it might be quite a substantial query for the database?)
> > > The cascade approach uses less code, and is pretty clear, but like you mentioned regarding language, I think it is a bit more risky on the performance side.
> > 
> > 
> > So I don't mind which is faster for deleting stuff, since that is a pretty rare occurence.
> > My concern is that adding relationships might make other unrelated things (that happen more often) slower - e.g. it looks like here if I get a `UserQuestionPublic` from the database I'll also now get a list with all the answers to that question, i.e. the same request now returns more data and is a bit slower? I'm not sure about this though, maybe the list is lazily fetched and it doesn't make any difference to the existing queries - I'll take a better look at this tomorrow.
> 
> This is the first time I've used a project with _Public for some returned calls, but it looks like it won't return answers by default to me. I could definitely be wrong though and that is absolutely something to be mindful about. I have just quickly debugged that by adding prints and deliberately failing the `test_get_user_question_admin_works` which returns the list of questions. That has UserQuestionPublic, but includes no answers, this isn't foolproof though:
> 
> image
> 
> The non-admin URI appears the same, answers aren't included: image
> 
> Nor are answers included in the OpenAPI schema: image

OK looks good! Thanks for checking this - I missed that `UserQuestionPublic` doesn't inherit from `UserQuestion`, so it doesn't have the `answers` attribute that was added to `UserQuestion`. The cascade approach looks like the way to go here then.

jmsssc:
Once this is merged I will again branch out to apply the correction scripts to this import_data directory ready for the 2nd batch of master theses import data

frontend/src/lib/components/DangerousDeleteModal.svelte
```diff
@@ -0,0 +1,121 @@
...
+			alertStore.showAlert(i18n.tr.admin.deleteError, "", true, false);
+			console.error(error);
+		}
+		if (data.would_delete) {
+			deletionWillAffectTotals = data.would_delete;
+		}
+	} else {
+		deletionWillAffectTotals = {};
+		deleteDone = false;
+	}
```
lkeegan, line 74: should we also clear the `deleteConfirmCode` here to avoid it being pre-filled next time the dialog is used?

frontend/src/lib/translations.ts
```diff
@@ -135,6 +135,18 @@ export const translationIds = {
 		editUserQuestionTitle: "Frage an Benutzer editieren",
 		expectedAgeData: "Voraussichtliches Alter Daten",
 		maxFileSizeIs: "Die maximale Dateigröße für ein Bild beträgt",
+		deletionWillAffect: "Durch die Löschung werden diese Elemente gelöscht:",
+		deletion: {
+			affectedQuestionAnswers: "Fragen Antworten",
+			affectedMilestones: "Meilsteinen",
+			affectedAnswers: "Meilsteinen Antworten",
+		},
```
lkeegan, line 143: the admin translation page assumes these are all `string:string` pairs, i.e. no nested objects, otherwise this breaks the admin translation page (I should have put a type def when defining this object to make this explicit, my bad!)

mondey_backend/src/mondey_backend/routers/admin_routers/milestones.py
```diff
@@ -69,11 +71,53 @@ def update_milestone_group_admin(
...
+                "deleted_answer_count": affected_milestone_answers,
+            }
+
+        affectedAnswers = 0
+        for milestone in milestone_group.milestones:
+            if milestone.id is None:
+                continue
+            affectedAnswers += count_milestone_answers_for_milestone(
+                session, milestone.id
+            )
```
lkeegan, line 112: here the logic to count affectedAnswers is duplicated in both branches - could this be done once before the `if dry_run` branch?

mondey_backend/src/mondey_backend/routers/admin_routers/milestones.py
```diff
@@ -112,11 +156,34 @@ def update_milestone(
...
+        dry_run: bool = Query(
+            True,
+            description="When true, shows what would be deleted without actually deleting",
+        ),
+    ):
+        affected_answers = count_milestone_answers_for_milestone(session, milestone_id)
+        if not dry_run:
+            milestone = get(session, Milestone, milestone_id)
+            if milestone is None:
+                return 0
```
lkeegan, line 171: This is not needed as this get utility function already raises a 404 in this case
```suggestion
```

mondey_backend/tests/routers/test_users.py
```diff
@@ -105,10 +113,117 @@ def test_create_update_and_delete_child(user_client: TestClient):
...
+    assert len(deleted_child_sessions) == 1
+
+    # Attempt to delete the child
+    response_delete = user_client.delete(
+        f"/users/children/{new_child_id}?dry_run=false"
+    )
+    assert response_delete.json()["deletion_executed"]
     assert len(user_client.get("/users/children/").json()) == 2
 
+    # Check that milestone answer sessions for this child have been deleted now
```
lkeegan, line 218: here i guess we should also check that ChildQuestions for this child have also been removed

mondey_backend/src/mondey_backend/routers/users.py
```diff
@@ -78,12 +81,32 @@ def update_child(
...
+                affectedAnswers += len(answering_session.answers)
+            return {
+                "ok": True,
+                "would_delete": {"affectedAnswers": affectedAnswers},
+            }
+
+        delete_statement = delete(MilestoneAnswerSession).where(
+            col(MilestoneAnswerSession.child_id) == child_id
+        )
+        session.execute(delete_statement)
```
lkeegan, line 106: we should also remove the child image if it exists here:
`child_image_path(child_id).unlink(missing_ok=True)`

lkeegan (changes requested):
Thanks - this looks good, I like the dangerously deleting modal, and now I'm much less worried about admins accidentally hitting deleting without realising the consquences!

It would be nice to also get the dangerously deleting modal when deleting questions in the admin interface.

Also when I click on a child as a user I see:

image

image

frontend/src/lib/components/DangerousDeleteModal.svelte
```diff
@@ -0,0 +1,121 @@
...
+			data: { deletion_executed: boolean; error?: never; would_delete?: never };
+	  }
+	| {
+			data: {
+				would_delete: DeletionWillAffectTotals;
+				error?: never;
+				deletion_executed?: never;
+			};
+	  }
+	| { error: string };
```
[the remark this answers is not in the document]
jmsssc: Absolutely I agree this is better. I was generating the openapi types but had some difficulties getting the response type to work without having a missing properties error. I will have a look at that.

jmsssc:
This pull request will close:

#273 - Delete with cascades for Milestones, Milestone Groups, User and Children questions

#321 - Adding option (and default hidden) for visibility for all questions

#333 - Bug when on the child page looking at a child, the  Questions to Children display included other child base data

mondey_backend/src/mondey_backend/models/questions.py
```diff
@@ -61,6 +63,7 @@ class UserQuestion(Question, table=True):
     answers: Mapped[list[UserAnswer]] = back_populates(
         "question", cascade="all, delete-orphan"
     )
+    visibility: bool = False
```
lkeegan, line 66: I'm not sure if we need to have visibility for individual texts - isn't it enough that the question is either visible or not?

mondey_backend/src/mondey_backend/models/questions.py
```diff
@@ -69,6 +72,7 @@ class UserQuestionPublic(QuestionPublic):
 
 class UserQuestionAdmin(QuestionAdmin):
     text: dict[str, UserQuestionText] = {}
+    visibility: bool = False
```
lkeegan, line 75: this inherits from QuestionAdmin which inherits from Question which already has the visibility member
```suggestion
```

mondey_backend/src/mondey_backend/models/questions.py
```diff
@@ -97,6 +102,7 @@ class ChildQuestionPublic(QuestionPublic):
 
 class ChildQuestionAdmin(QuestionAdmin):
     text: dict[str, ChildQuestionText] = {}
+    visibility: bool = False
```
lkeegan, line 105:
```suggestion
```

mondey_backend/src/mondey_backend/models/questions.py
```diff
@@ -89,6 +93,7 @@ class ChildQuestion(Question, table=True):
     answers: Mapped[list[ChildAnswer]] = back_populates(
         "question", cascade="all, delete-orphan"
     )
+    visibility: bool = False
```
lkeegan, line 96:
```suggestion
```

mondey_backend/src/mondey_backend/routers/questions.py
```diff
@@ -19,15 +19,19 @@ def get_user_questions(
...
         user_questions = session.exec(
-            select(UserQuestion).order_by(col(UserQuestion.order))
+            select(UserQuestion)
+            .where(UserQuestion.visibility)
+            .order_by(col(UserQuestion.order))
         ).all()
 
         return user_questions
 
     @router.get("/child-questions/", response_model=list[ChildQuestionPublic])
```
lkeegan, line 29: could we include an invisible question in the backend tests? i.e. add one to conftest, then check that it is only included in the admin questions endpoint and not the user one.

mondey_backend/src/mondey_backend/models/questions.py
```diff
@@ -44,6 +45,7 @@ class QuestionPublic(SQLModel):
     additional_option: str = ""
     required: bool = False
     name: str = ""
+    visibility: bool = False
```
lkeegan, line 48: since the backend only returns `QuestionPublic` objects if they are visible, I don't think we need to include this here - visibility is not something the frontend should care about, it just displays whatever the backend gives it
```suggestion
```

lkeegan (approved):
thanks - some minor comments, otherwise lgtm!

mondey_backend/src/mondey_backend/models/questions.py
lkeegan: This comment is wrong (sorry I got fooled by github's code folding), but I think we can still remove visibility since this inherits from Question which has it already

lkeegan (approved):
thanks - lgtm!

Issue #307, opened 2025-04-07
Issues with imported questions


lkeegan:
- [x] german options text using `,` instead of `;` to separate options
  - can be fixed in admin interface
- [x] options list missing
  - can be fixed in admin interface (just copy german text)
- [x] english texts missing from database
  - needs to be fixed by adding these to the database, then editing them in the admin interface
- [x] some junk in the text entries like `<tab>` and `&#44;`
  - if this is fixed/edited we need to also edit the existing answers in the database to match
  - I replaced `&#44;` with `,` in the questions in the admin interface
    - need to check if we also need to do the same in the corresponding database answers
    - `;` is used to separate the options

lkeegan:
FP01-FP05 seem like a two-level "relation to child" question - i.e. first they answer "Eltern", and then they give further info by answering the corresponding other question - is that correct?

And if so can we convert this to a single longer drop-down select question to fit into the supported types of questions?

![Image](https://github.com/user-attachments/assets/13f6288c-6849-4c37-892f-397f972fb3f3)

lkeegan:
ChildQuestions FK01 and FK02 can be removed (Birth year&month, since they are intrinsic properties of the `Child`)

jmsssc:
Okay, so I merged the two level question and am adding Andere Dianogsen answers to Gesundheit Andere and deleting Andere Diagnosen.

For the Health issues: Right now they do show as a two option select (nicht gewählt;ausgewählt). Should I convert them to Ja/Nein on the frontend or something else?

![Image](https://github.com/user-attachments/assets/0b767a30-8701-431d-94e1-2e65a192df13)

lkeegan:
> Okay, so I merged the two level question and am adding Andere Dianogsen answers to Gesundheit Andere and deleting Andere Diagnosen.
> 
> For the Health issues: Right now they do show as a two option select (nicht gewählt;ausgewählt). Should I convert them to Ja/Nein on the frontend or something else?
> 
> ![Image](https://github.com/user-attachments/assets/0b767a30-8701-431d-94e1-2e65a192df13)
> 

Great! yes Ja/Nein sounds good to me

lkeegan:
All questions now flattened and approved by researchers

PR #312, opened 2025-04-08
Fix missing additional_answers in research data

### LINKED ISSUE #308: aditional_answers missing from research data
**Describe the bug**
Currently the research data just provides `answer` for all answers, but if that answer is the equivalent of "other" and there is an additional free-text answer provided, we should use that additional answer instead.

**Context**
A `UserAnswer` or `ChildAnswer` has an `answer` and an `additional_answer`.

If the corresponding question specifies an `additional_option`, then when the user selects that option, a free text entry box is provided. In this case `answer` is the value of `additonal_option` (e.g. Andere) and `additional_answer` contains their free text answer.

**Solution**

Backend needs to look-up the question for each answer, and if `answer` matches the `additional_option` in the question then we should use the `additional_answer` instead of `answer` in the research data

lkeegan:
- if the `answer` to a question is the `additional_option` from the question
  - the user entered a free text answer `additional_answer`
  - this `additional_answer` is now used instead of `answer` in the research/data endpoint
- resolves #308

### DIFF
--- a/mondey_backend/src/mondey_backend/statistics.py
+++ b/mondey_backend/src/mondey_backend/statistics.py
@@ -522,36 +522,45 @@ def get_milestone_answers(session: SessionDep) -> pd.DataFrame:
     return df
 
 
-def get_user_answers(session: SessionDep) -> pd.DataFrame:
-    user_question_ids = session.exec(
-        select(UserQuestion.id).order_by(col(UserQuestion.id))
-    ).all()
-    user_answers: dict[int, dict[str, str | int | float]] = defaultdict(dict)
-    for answer in session.exec(select(UserAnswer)).all():
-        user_answers[answer.user_id][f"user_question_{answer.question_id}"] = (
-            answer.answer
+def get_answers(
+    session: SessionDep,
+    answer_type_name: str,
+) -> pd.DataFrame:
+    if answer_type_name == "user":
+        question_type = UserQuestion
+        answer_type = UserAnswer
+    elif answer_type_name == "child":
+        question_type = ChildQuestion  # type: ignore
+        answer_type = ChildAnswer  # type: ignore
+    else:
+        raise ValueError(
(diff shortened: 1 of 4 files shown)


mondey_backend/src/mondey_backend/statistics.py
```diff
@@ -522,36 +522,45 @@ def get_milestone_answers(session: SessionDep) -> pd.DataFrame:
...
+    question_additional_options = {
+        q.id: q.additional_option
+        for q in session.exec(
+            select(question_type).order_by(col(question_type.id))
+        ).all()
+    }
+    answers: dict[int, dict[str, str | int | float]] = defaultdict(dict)
+    for answer in session.exec(select(answer_type)).all():
+        answer_value = answer.answer
+        # if the selected answer is the additional option, use the additional answer instead
```
jmsssc, line 548: Looks great to me!

I am not sure if this will work with the imported data since the import process did not set the additional_option column value for those questions with an additional option. I can fix that quickly in the import script branch, and/or I can grab the question IDs/respective tables to update that for the import?

There are two kinds of free text answer that were imported:

Some, like "Fruhgeboren [01]", simply function as freetext and relate to the preceding question, but should not overwrite the categorical answer to the previous question

Others, like Sprachen, are freetext and set the "additional_answer" property, leaving the answer property to a value like "Andere" or "Other".

I also noticed a small issue on the research page: No graph data between 6 and 12 months seems to appear? It might be related to my database data?

image

image


(this is with the imported data)

This appears to be an issue at least as far back as commit:
b495b6d6751d35cb68dc390dc62a7a68eb886c29

and possibly earlier to the research page initial integration?

lkeegan: in general there aren't many data points for the younger kids - many of the points on these plots are averaged over only a couple of points:

image

But yes absolutely we/they should check that the data agree!

Regarding the `additional_answer` let's discuss in person what the best way to proceed might be - I've made some notes in #307

PR #330, opened 2025-04-10
302 minor styles todos

jmsssc:
Small improvements based on Igors feedback, fixes no bugs, only changes the style.
* Also removes image upload functionality and rendering images of children in the UI by a feature boolean we can re-enable later


frontend/src/routes/userLand/children/gallery/+page.svelte
```diff
@@ -14,6 +14,7 @@ import { alertStore } from "$lib/stores/alertStore.svelte";
 import { currentChild } from "$lib/stores/childrenStore.svelte";
 import { type CardElement, type CardStyle } from "$lib/util";
 import { Heading, Spinner } from "flowbite-svelte";
+import { displayChildImages } from "../../../../features";
```
lkeegan, line 17: Could we put the features.ts in lib and use an absolute import? Makes this import less brittle than a relative one if we refactor these files later
```suggestion
import { displayChildImages } from "$lib/features";
```

lkeegan (approved):
one minor comment - also does the month localization work for you? for me locally it's always english, although the code looks like it should work to me.

otherwise lgtm!

jmsssc:
> one minor comment - also does the month localization work for you? for me locally it's always english, although the code looks like it should work to me.
> 
> otherwise lgtm!

It did work for me (in German), but I will have a look! I added "en" as  a back up but could change the code and look at that. If you change to hardcode it to e.g. "de" maybe then it would change? (and if not, it's definitely not working)

lkeegan:
> > one minor comment - also does the month localization work for you? for me locally it's always english, although the code looks like it should work to me.
> > otherwise lgtm!
> 
> It did work for me (in German), but I will have a look! I added "en" as a back up but could change the code and look at that. If you change to hardcode it to e.g. "de" maybe then it would change? (and if not, it's definitely not working)

I think this might have been a caching issue for me locally, or just a mistake on my side, in any case this is working in production :-)

PR #331, opened 2025-04-10
Improve additional answer part of admin question UI

### LINKED ISSUE #313: improve admin question interface

The "additional option" interface is a bit clunky. This is used to make one of the options the "other" option where the user can enter a free text response.

- currently the admin has to manually type the option that should have this behaviour
- the preview widget on the right doesn't implement the additional option behaviour

Image

Instead of typing the option, there should be a checkbox to enable/disable free text entry.

If enabled, then the free text option should be selected from a dropdown of the available options instead of being typed manually.

Also the preview widget should implement this free text behaviour.

lkeegan:
- selection of free text option now down using dropdown
- preview now includes free text functionality
- resolves #313


frontend/src/lib/components/Admin/EditQuestionModal.svelte
```diff
@@ -37,6 +37,18 @@ let {
 } = $props();
 let preview_lang = $state("de");
 let preview_answer = $state("");
+let options = $derived.by(() => {
+	let opts = [{ value: "", name: "No free text option" }];
+	if (!question || !question.options) {
+		return opts;
+	}
+	return opts.concat(
+		question.options.split(";").map((value) => ({
```
jmsssc, line 46:
```suggestion
		question.options.split(";").map((value) => ({
```

```suggestion
		question.options.replace(/;$/, '').split(";").map((value) => ({
```

In order to prevent a trailing ";" creating a fake option

image

jmsssc (approved):
Looks good to me and the preview works. If the user is a bit of a programmer and adds a trailing ";" to the raw options IDs, it can create fake/empty options in the Select which cannot be selected correctly, so maybe we should strip that when they edit the select questions?

image

Issue #326, opened 2025-04-10
Raw data research download button


lkeegan:
Add a button to download the full raw data (i.e. just the /research/data endpoint as a csv file) to the research page.

jmsssc:
Completed in https://github.com/ssciwr/mondey/pull/337

PR #337, opened 2025-04-15
Complete #326 - Add button that downloads /research/data endpoint as CSV

jmsssc:
I added this functionality. Right now the download all contains the data from research/data which means it does not include the Child CASE numbers. Should we maybe include those so that if later it is identified that e.g. CASE 23 needs to be removed (because consent is withdrawn or something) that they can remove that data from the CSV they are using?

Image

Though, with our recent changes in #273, it's possible to simply delete the child from the interface instead, which in my eyes is actually preferable.


lkeegan (changes requested):
thanks - this looks good, but I think we could do this is a much simpler way without any backend changes:

- the web worker already constructs `df_in` which is a danfo.js dataframe with the full data
- so [df_in.toCSV({download: true})](https://danfo.jsdata.org/api-reference/input-output/danfo.to_csv) should download the csv
- (might have to check if the download bit works inside a webworker though)

jmsssc:
> thanks - this looks good, but I think we could do this is a much simpler way without any backend changes:
> 
>     * the web worker already constructs `df_in` which is a danfo.js dataframe with the full data
> 
>     * so [df_in.toCSV({download: true})](https://danfo.jsdata.org/api-reference/input-output/danfo.to_csv) should download the csv
> 
>     * (might have to check if the download bit works inside a webworker though)

Yes it would be good to do this clientside anyway. Setting all columns & milestones to be selected would be a (too) hacky way.

When running that code in the worker I run into 
image

So to me it sounds like we could achieve this on the clientside without reloading with a boolean flag to the worker which when true returns the CSV full data of df_in, and then on the frontend, we download it manually as in the existing function for the selected/partial milestones and columns. I think returning a DF might not work because of having to encode it into a basic format to send in a webworker onMessage.

What do you think?

lkeegan:
> > thanks - this looks good, but I think we could do this is a much simpler way without any backend changes:
> > ```
> > * the web worker already constructs `df_in` which is a danfo.js dataframe with the full data
> > 
> > * so [df_in.toCSV({download: true})](https://danfo.jsdata.org/api-reference/input-output/danfo.to_csv) should download the csv
> > 
> > * (might have to check if the download bit works inside a webworker though)
> > ```
> 
> Yes it would be good to do this clientside anyway. Setting all columns & milestones to be selected would be a (too) hacky way.
> 
> When running that code in the worker I run into image
> 
I guess you need access to the DOM to do the download, which the worker doesn't have.

> So to me it sounds like we could achieve this on the clientside without reloading with a boolean flag to the worker which when true returns the CSV full data of df_in, and then on the frontend, we download it manually as in the existing function for the selected/partial milestones and columns. I think returning a DF might not work because of having to encode it into a basic format to send in a webworker onMessage.
> 
> What do you think?

Yes sounds good (I would add a new message type to the worker for this)

jmsssc:
I converted this to draft because after checking this morning again, I'm going to add a loading indicator to the download button as if the data became much larger the user might think it was broken while it loads @lkeegan

jmsssc:
Just a note to myself:

The pre-commit nesting } error on app.css was because without a terminating semi-colon for CSS with @apply, pre-commit processes it differently to the app (which will work fine without semi-colons, which are optional for a single line in a class in CSS in practise by browser behavior). The fix is to use a terminating semi-colon on the end of each css line. It is confusing, because everything will work fine and CSS will apply without the semi-colon, but it throws off the pre-commit parser.

jmsssc:
@lkeegan this is ready for review now

lkeegan (changes requested):
thanks - looks good, I like the worker type enums!

I don't think we need to display the full table of data in the browser though - this is already very slow to render and will get worse with more data, and not sure how useful it is.

Also if we skip displaying the json then we can just directly return csv_data from the webworker with danfo.toCSV instead of doing df -> json in the worker and then json -> csv in the frontend. This should also make the download pretty much instantaneous.

jmsssc:
> thanks - looks good, I like the worker type enums!
> 
> I don't think we need to display the full table of data in the browser though - this is already very slow to render and will get worse with more data, and not sure how useful it is.
> 
> Also if we skip displaying the json then we can just directly return csv_data from the webworker with danfo.toCSV instead of doing df -> json in the worker and then json -> csv in the frontend. This should also make the download pretty much instantaneous.

Sure this makes sense to me! Especially leaving json_data alone and not changing it.

Just to check, would you use danfojs to convert to CSV and send that data to frontend as a string. Then turn that string into a simple file we download with something like a javascript-initiated button click (a bit hacky/2010s style)... e.g.
```
const blob = new Blob([csvString], { type: 'text/csv;charset=utf-8;' });
const url = URL.createObjectURL(blob);
const link = document.createElement('a');
  link.href = url; // our CSV as data object URL
  link.setAttribute('download', filename);
  document.body.appendChild(link);
  link.click();
```
That's just because the export-to-csv library doesn't support simple string CSVs as inputs for download, but rather formats like JSON. But that's an awful lot of extra kbs of string keys indeed to send over.

I think another library would support simple CSV string download but I wanted to check your thoughts first before proceeding, as the javascript element+click approach seems a bit hacky for 2025.

jmsssc:
n.b. to self: need to encode the commas before running toCSV... the commas in the bracketed examples (e.g. row 20) cause the data to lose alignment...
image

frontend/src/routes/userLand/research/+page.svelte
```diff
@@ -126,13 +159,25 @@ function downloadCSV() {
...
+	is_downloading = false;
+}
+
+function handleDownloadAll() {
+	if (downloadAllHandler) {
+		is_downloading = true;
+		downloadAllHandler();
+		// Reset state
+		json_data = [];
+		plot_data = {} as PlotData;
```
lkeegan, line 179: now that downloading doesn't modify json_data I guess we don't need to reset this:
```suggestion
```

frontend/src/routes/userLand/research/+page.svelte
```diff
@@ -4,6 +4,8 @@
 import PlotLines from "$lib/components/DataDisplay/PlotLines.svelte";
 import { i18n } from "$lib/i18n.svelte";
 import { type PlotData } from "$lib/util";
+import pkg from "file-saver";
+const { saveAs } = pkg;
```
lkeegan, line 8:
```suggestion
import saveAs from "file-saver";
```

lkeegan (approved):
thanks - looks good! made a few very minor suggestions. and good catch on the quoting of the csv!

PR #338, opened 2025-04-15
336 note lack of data on feedback view

jmsssc:
This changes the UI to let users know when there is absolutely no peer data for a child with milestone answers.

I am not sure we need to make this change after trying out the UI, I know our conclusion in the early April meeting was not super one way or the other.

Basically, if children lack peer data, an invalid value will come through on the UI saying "Not enough total data" / "Noch nicht genügend Daten für Feedback".

If there is any valid data, that won't happen, but I think in that case it is fine.

The situation that caused me to be confused as to how adding milestone answers with value "0" I am nolonger confused about and it functioned as it should have in that circumstance.

@lkeegan do you have any thoughts?


jmsssc:
Closing as https://github.com/ssciwr/mondey/pull/341 will solve this in tune with the intentions of the Mondey feedback intention

PR #341, opened 2025-04-22
Don't provide feedback if too few datapoints, and use 0-3 scale for s…

### LINKED ISSUE #271: Score scale 0-3 or 1-4?
We use 0-3 for the answers in the database, and convert to 1-4 in various places.
However in the sample data 0-3 is also used, so maybe we don't need to convert to 1-4?

If possible / desired, we could just use 0-3 everywhere.

### LINKED ISSUE #336: Feedback page should note in grey when there are insufficient children at similar age for comparison
As discussed in #305 and then in our early April Mondey meeting.

**Currently**
- Children who have no/limited data see a grey note that a comparison for milestones cannot be made.

**In the future**
- The same previous note should remain.
- For children who **have data, but not enough peers for comparison**, an additional gray note (appearing above and not replacing the green ticked milestones for any milestones the child has been marked in any form with data upon). Backend can send back a special param for when this is the case, and a grey sentence should be rendered above the milestones.

lkeegan:
…cores

- don't provide feeback if too few datapoints
  - add `min_num_samples` parameter to `_finalize_statistics`
    - if there are fewer than this number of samples, we set the standard deviation to -1
    - currently this is set to 2 for testing so we see some feedback with the current data
    - can increase this later in production
  - compute_feedback_simple
    - now returns (existing) `invalid` trafficlight if standard deviation is negative
  - resolves #336
- use 0-3 for answer score scale
  - reduces possibility of bugs due to converting between 0-3 and 1-4 scales in various places
  - NOTE: after merging this we need to call the `/admin/update-stats/false` endpoint to update all statistics in the database
  - resolves #271

### DIFF
--- a/frontend/src/lib/components/DataDisplay/PlotLines.svelte
+++ b/frontend/src/lib/components/DataDisplay/PlotLines.svelte
@@ -42,10 +42,10 @@ onMount(() => {
 				label: `${i18n.tr.admin.age} (m)`,
 				tickValues: [1, 2, 3, 4, 5, 6, 9, 12, 16, 24, 36, 48, 60, 72],
 			}),
-			yAxis: new Axis({ label: `${i18n.tr.admin.averageScore} (1-4)` }),
+			yAxis: new Axis({ label: `${i18n.tr.admin.averageScore} (0-3)` }),
 			xScale: Scale.scalePow().exponent(0.5),
 			xDomain: [1, 72],
-			yDomain: [1, 4],
+			yDomain: [0, 3],
 		},
 		plot_data.data,
 	);
--- a/frontend/src/lib/components/DataDisplay/PlotScoreAge.svelte
+++ b/frontend/src/lib/components/DataDisplay/PlotScoreAge.svelte
@@ -36,14 +36,16 @@ onMount(() => {
 					color: "#000000",
 				}),
 			],
-			xAxis: new Axis({
+			xAxis: new Axis<MilestoneAgeScore>({
 				label: `${i18n.tr.admin.age} (m)`,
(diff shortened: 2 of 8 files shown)


jmsssc (approved):
This seems all good to me although I do notice two things:
A) Young children without much data will see a lot of Yellow warning signs in milestone groups where they have no data (and no note) which I believe is intended behavior.

B) In the recent feedback email, the following was stated which differs from our current conception of the feedback page ... "The user would probably need a hint saying that feedback can only be provided once milestones for all domains have already been evaluated (maybe this hint could be located underneath the list of milestones for each domain?" indicates to me  maybe there is some difference in understanding, because we do provide feedback when there is really any amount of data for a child and their is peer data in quantity. It does not affect this PR per say but is related. We might want to hide all feedback if not all milestones are filled in for the child on the feature page as a fix for this. We should double check that is what is meant though on the 5th.

lkeegan:
> This seems all good to me although I do notice two things: A) Young children without much data will see a lot of Yellow warning signs in milestone groups where they have no data (and no note) which I believe is intended behavior.
> 
> B) In the recent feedback email, the following was stated which differs from our current conception of the feedback page ... "The user would probably need a hint saying that feedback can only be provided once milestones for all domains have already been evaluated (maybe this hint could be located underneath the list of milestones for each domain?" indicates to me maybe there is some difference in understanding, because we do provide feedback when there is really any amount of data for a child and their is peer data in quantity. It does not affect this PR per say but is related. We might want to hide all feedback if not all milestones are filled in for the child on the feature page as a fix for this. We should double check that is what is meant though on the 5th.

Yes we're supposed to only give feedback when all milestones have been answered for a session, I need to look into what we actually do here.. in general I think there are likely a lot of minor things to fix with the feedback logic as we still haven't really tested this properly yet: https://github.com/ssciwr/mondey/issues/347

PR #349, opened 2025-04-24
Fix translations not being updated

### LINKED ISSUE #348: Translations not being saved
**Describe the bug**
When I add a translation in the admin interface & click save it doesn't give any error messages (neither frontend nor backend), but the translated text is not updated.

lkeegan:
- add `no-cache` to fetch request when getting i18n json files from server in admin interface
  - otherwise we don't see the changes made until whenever the cached version is invalidated
  - resolves #348
- disable auto-translate button after it is clicked to avoid spamming deepl api with requests

### DIFF
--- a/frontend/src/lib/components/Admin/InputAutoTranslate.svelte
+++ b/frontend/src/lib/components/Admin/InputAutoTranslate.svelte
@@ -1,11 +1,16 @@
-<svelte:options runes={true} />
+<svelte:options runes={true}/>
 
 <script lang="ts">
 import { translate } from "$lib/client";
-import { i18n } from "$lib/i18n.svelte";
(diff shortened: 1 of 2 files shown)


jmsssc:
I am reviewing this now

jmsssc (approved):
This looks good to me. I deleted the previous deepl key so please send me it again if you would like me to run this with the key, but the changes here look contained, don't cause the local version to crash, and make sense to me to avoid the caching issue while adding some spam protection from Deepl calls.

jmsssc:
@lkeegan Yes this works for me. Requesting multiple translations (by opening a group in the translations admin page with several missing and clicking the magic wand for 4-5 at once) works and they slowly load in one-by-one. One at a time, it takes at most a few seconds. All good to merge.

PR #350, opened 2025-04-28
Automatically identify suspicious answer sessions

### LINKED ISSUE #297: Expose summary of data to admins/researchers
It would be nice to provide some summary for the admins of the data in the system,
as well as to potentially identify outliers / junk data / inconsistencies.

- counts of
  - users
  - children
  - answersessions
  - answers
- the same thing but as a function of time, e.g. binned by month?
- the same thing but as a function of child age (excl. users)?
- for each child/user question, a list of all answers with the frequency/count for each
- sum of children vs number of answersessions for that child

Unclear if this should be in admin or research.

### LINKED ISSUE #323: Identify outliers automatically
System should identify child answers that are many sd away from expected values.

These answer sessions should be flagged with a `suspicious` flag.

Any suspicious answer sessions should be excluded from the statistics calcs for feedback / expected age calculations.

However they should still be included in the research data.

Also an admin should be able to set the suspicious flag on/off.

lkeegan:
- backend
  - add `suspicious` flag to answer session
  - add `MilestoneAnswerAnalysis` and `MilestoneAnswerSessionAnalysis` models
    - contains rms of difference between answers in a milestone answer session and the average answers for that age
    - also the individual answers and average answers for each milestone in the session
  - add endpoints
    - `/admin/milestone-answer-sessions/`
      - provides a list of all answer sessions
    - `/admin/milestone-answer-session-analysis/{answer_session_id}`
      - provides analysis results for a single answer session
    - `/milestone-answer-sessions/{answer_session_id}?suspicious={true/false}`
      - allows admin to set suspicious flag value for an answer session
  - call `flag_suspicious_answer_sessions` when calculating statistics
    - this sets the `suspicious` flag for any sessions with an rms difference from average for that age > 1.0
    - only applied to sessions that are not yet included in the statistics
  - exclude suspicious answer sessions when calculating statistics, but still include them in the research data
  - when re-calculating all statistics, set `included_in_statistics` to false for all sessions initially
    - otherwise if a session that was previously included in the stats is marked suspicious by an admin its `included_in_statistics` value would never get updated
  - update e2e sql data & conftest accordingly
    - add missing milestoneanswersession to e2e sql file
- admin-frontend
  - add Data tab
    - lists all the milestone answer sessions
    - add checkbox to show only suspicious answer sessions
  - add AnswerSessionAnalysisModal
    - show plots of answers & difference to average answers, along with child age and rms difference
    - allow admin to mark session as suspicious or not
- resolves #323
- resolves #297


jmsssc (changes requested):
# Functionality of Daten Page
This works well for me on first use, I can understand which sessions are possible outliers, suspicious, get a useful overview of the answers vs the average answers, and manually set suspicious sessions. They don't influence or change on the research page as expected which is great. I can mark and edit the suspicious flags and get it to identify them for me with calculate all. 

# Functionality of hiding suspicious answers on feedback page
When I manually set all answering sessions to suspicious, feedback disappears, as I expect, with one exception[1]. I maybe missed it, but where does the backend filter for the suspicious sessions for the API routes that the feedback page uses? e.g, this route still returns suspicious ones which get through on the feedback page for a specific child:

``` 
getExpiredMilestoneAnswerSessions({
		path: { child_id: currentChild.id as number },
	})
```

[1] Even though I set this one to suspicious (via SQL execution, I did it for all answer sessions), it still makes its way to the frontend feedback page for this child when I debugged:
image

image


image


# Interface
This looks great to me, and in terms of usability maybe we should expand RMS to Root Mean Squares*, and for the graphs, add a horizontal line to clarify that the above is the points for this child, and the one below is the milestones relative to the average score - it works really well how they align & match so you can look between the two and the milestones are aligned, just the y axis labels are a bit tricky to comprehend at first, maybe a title or horizontal bar can clarify these graphs, or maybe something like a header saying "Comparison to average child score" above the second graph?

We could also have a note pop up saying "Statistics out of date: Update now" or something when they set some suspicious ones to non-suspicious (to entice them to click "Statistiken aktualisieren").

I would also suggest we clarify on the page with a description that "suspicious sessions are included" on the research page and on the Data page, incase the user marks them as suspicious intending them to be removed from the research page.

lkeegan:
Thanks for reviewing - I've implemented the interface suggestions in 92a39ba

Re functionality - a user should still see their answer session even if it is suspicious. The only effect of the suspicious flag is that any suspicious answer sessions are ignored when we are calculating the expected ages / average scores for milestones.

jmsssc (approved):

Issue #346, opened 2025-04-23
Able to proceed: Late April UI change feedback


jmsssc:
### Milestone Redirect
- [x] Done
Once all milestones of a given domain have been evaluated, the system should automatically proceed to the overview-milestone page (all domains) again, so that the user does not need to click this header again.

**Currently** It redirects to the milestone group the user just filled out, showing the milestones they just specified:

![Image](https://github.com/user-attachments/assets/b15cb8e7-28ad-4816-bf13-2068a05baf53)

**Change** It should redirect to the milestone overview:

![Image](https://github.com/user-attachments/assets/328680a8-1b89-4345-b198-a55ac360eac2)

### Milestone skip option
- [x] Done
It would be good if we did not only have the evaluation buttons (noch gar nicht, in Ansätzen, Weitgehend, Zuverlässig) but also one saying “später bewerten” / “evaluate later” for cases in which the observer still needs to check whether the child can do it.

**Currently** The user has to answer each milestone in essence, when they start to "score" a milestone group.

![Image](https://github.com/user-attachments/assets/dafb02d8-3839-4f01-952f-bf36511ef4b9)

**Change** Un-disable the "forward" button and make it read Evaluate Later. It should not record a value for this milestone score answering session.

### Kinder Search Field width
- [x] Done
“Kinder nach Namen durch….“ Is it possible to adjust the size of the search button so that the explanation “Kinder nach Namen durchsuchen” is fully visible?

**Currently:** Width too small:

![Image](https://github.com/user-attachments/assets/ff1e6550-79c0-43cd-a036-40b42848608c)

### Terminology changes
- [x] Ausprobieren Tooltip:  Once your cursor hovers over this field a short explanation should pop up, saying that “You can try out the program, but your data won’t be stored and cannot be used for providing feedback.” (German: “Sie können das Programm ausprobieren, ohne dass Ihre Daten gespeichert werden; sie lassen sich dann aber auch nicht für die Feedback-Funktion nutzen“). That way, the user know what will happen before pressing this button.
- [x] Please replace „Benutzerdaten eingeben“ by “Daten zur eigenen Person eingeben“ / Enter user information” (to avoid confusion whether the observer should enter his/her own data or that of the child).
- [x] “Wählen sie ein Kind zur Beobachtung aus oder legen melden sie ein neues Kind an“ The word „melden“ should be cut out.
[...]

PR #352, opened 2025-05-04
346 quick changes

jmsssc:
Changes to complete #346


jmsssc:
Not ready yet sorry. The favicon showed as a green checkmark so I thought the tests passed (as they did locally), but apparently not.

frontend/src/routes/userLand/milestone/+page.svelte
```diff
@@ -199,8 +200,22 @@ const breadcrumbdata = $derived([
                                 class={` min-w-20 absolute top-0 left-0 w-full h-full object-cover transition duration-1000 ease-in-out ${imageIndex === currentImageIndex ? 'opacity-100' : 'opacity-0'}`}
                                 src={`${import.meta.env.VITE_MONDEY_API_URL}/static/m/${image.id}.webp`}
                                 alt=""
+                                on:click={() => {
+                                    modalImagePreviewOpen = true;
+                                }}
                         />
                     {/each}
+                    <Modal bind:open={modalImagePreviewOpen} size="xl" autoclose outsideclose>
+                        <div class="p-2">
```
lkeegan, line 209: on (emulated) mobile I struggled to close this modal, as the close button was hidden under the navbar (due to its high z index) and clicking outside didn't seem to work. Do you see the same issue?

image

Maybe we could also make this modal close when it is clicked on, e.g.?
```suggestion
                        <div class="p-2" on:click={()=>{modalImagePreviewOpen=false}}>
```

frontend/tailwind.config.ts
```diff
@@ -69,6 +70,7 @@ export default {
 					1: "#cdd993",
 					2: "#a7c957",
 					3: "#6a994e",
+					"-1": "#274e88",
```
lkeegan, line 73: For me it's a bit hard to read the mouseover text on this blue, maybe the blue should be a bit lighter?
image

mondey_backend/src/mondey_backend/routers/users.py
```diff
@@ -63,6 +65,13 @@ def create_child(
         current_active_user: CurrentActiveUserDep,
         child: ChildCreate,
     ):
+        current_month = int(datetime.now().strftime("%m"))  # 01 for Jan
+        current_year = int(datetime.now().strftime("%Y"))
+        if (
+            child.birth_month > current_month and child.birth_year >= current_year
+        ) or child.birth_year >= current_year:
```
lkeegan, line 72: here we can avoid doing the logic (which is easy to get wrong) ourselves by comparing the dates:
```suggestion
        if datetime(child.birth_year,child.birth_month,1) > datetime.now():
```

lkeegan (approved):
thanks looks good! made a couple of minor suggestions, also I pushed a commit to make sure backend & frontend stayed in sync when the 'answer later' button is used.

mondey_backend/src/mondey_backend/routers/users.py
jmsssc: Sounds good

Issue #356, opened 2025-05-05
Change icons on Milestones (NOT feedback page) to full circles


jmsssc:
As discussed via email:

![Image](https://github.com/user-attachments/assets/3ce157e6-828a-4e71-a13d-69ccc07a1744)

PR #360, opened 2025-05-07
#356 Make milestone checkmarks/exclamation points into (solid-able) w…

jmsssc:
…hite circles. Need to check where else the auxilarry part of CardDisplay is used and that it has not changed those parts.


jmsssc:
image

lkeegan (approved):
lgtm!

PR #362, opened 2025-05-08
Remove paragraph from milestones + Make 2-3 columns appear depending on screen size

jmsssc:
Yes we said 4 was okay on big screens, but I think the breakpoint for big is much smaller than we expect - it's essentially not much bigger than a laptop screen). Plus fix the search menu being broken for cases where there is search data.


lkeegan (approved):
lgtm!

PR #363, opened 2025-05-09
Fix milestone group svg/stddev calculation

### LINKED ISSUE #174: Implement final algorithm for expected ages
We have a simple placeholder algorithm for this for now: average the score for each age month, pick the first month with an average score of 3/4 or greater.

https://github.com/ssciwr/mondey/blob/85aadd41eb2e9205140c470d9e872a3a3c74029b/mondey_backend/src/mondey_backend/routers/utils.py#L180-L223

Once the website has been populated with the existing data from previous surveys (and possibly some new user data), we should revisit this algorithm and implement the final algorithm to be used in production.

### LINKED ISSUE #347: When should feedback be provided?
Our understanding:

- when a user starts answering milestones this starts a "milestone answer session"
- they have a week to answer all of the provided milestones
- after a week the session expires, and if they go to answer milestones they start a new session
- once they have answered all the questions in a session, feedback should be calculated and displayed
- if they didn't answer all the questions and the session has expired, no feedback should be provided

Todo:

- confirm what we actually do in the code currently
  - the answer sessions are implemented as above
  - I'm not sure exactly what we do for the feedback
- confirm that the above is what is supposed/desired to be done

### LINKED ISSUE #359: Fix Milestone group avg/sd calculation
When a user answers the milestones in a milestone group, they only provide answers for the subset of milestones with expected age in the range child age +/- 6 months.

However, in all the imported data they answered *all* the milestones, regardless of expected age.

Currently we provide milestone group feedback based on the average and std. dev. of these scores, but this is not a fair comparison as we are comparing different sets of milestones.

(There is also the issue that the set of milestones may change over time as the expected age gets updated, but hopefully that will not result in a lot of changes)

We need to compare apples with apples, i.e. we should use the same set of milestones when calculating avg/std dev as when calculating the values to compare to these, which can be dones in two ways:

1. always include all milestones, setting a score of 0 for all older unasked milestones and a score of 3 for all younger unasked milestones before calculating the avg/std dev
2. only use the milestones from the answer session for which we are calculating feedback

I think 1 is the better approach, as 2 would require re-calculating the statistics for every feedback session and would also be more sensitive to the milestone expected age changing.

lkeegan:
- include all milestones in a group when calculating score avg/std dev for that group
  - if a milestone was not included in a milestone answer session
    - assume a score of 3 if the child is older than the milestone expected age
    - assume a score of 0 if the child is younger than the milestone expected age
  - this ensures we are comparing avg/std dev across the same populations
- update and extend tests
  - include stats session into main test session
    - otherwise makes it harder to ensure the pre-constructed statistics are correct as they depend on which answer sessions are present
- resolves #359

Improve how answer session progress is tracked and communicated

- add `completed` field to `MilestoneAnswerSession`
  - completed flag is set when the last remaining milestone answer for a session is provided
    - the expired flag is also set at the same time to prevent a completed session from being modified
    - remove no longer needed code to expire stale answer sessions when calculating stats - now we just filter on completed answer sessions
  - only completed sessions are used to calculate averages and feedback
  - user feedback is now only provided for completed answer sessions
  - response from milestone-answers endpoint now includes `session_complete` bool so that frontend can redirect to feedback page when answer session is complete
- add `active_answer_session`, `session_progress`, `session_remaining_seconds` for each child to `/children` endpoint
  - frontend uses these to display progress bar and time remaining for the current session for each child in overview
  - add luxon library to convert duration in seconds to localised text
- change answer session duration to 14 days
- resolves #347

Also

- update algorithm to calculate expected age based on current data
  - require at least 3 answers for each age
  - expected age is the first age where average score >= 2.1 (where 0.0 is the minimum score and 3.0 is the maximum)
  - resolves #174
- remove `expected_score` from `MilestoneAgeScore`
  - was only used in frontend in `PlotScoreAge`
  - now inferred in frontend from `expected_age` in `MilestoneAgeScoreCollectionPublic`


frontend/src/lib/components/DataDisplay/CardDisplay.svelte
```diff
@@ -13,87 +15,111 @@ let {
...
+		return "";
+	}
+	const duration = Duration.fromObject(
+		{ seconds: data.secondsRemaining },
+		{ locale: i18n.locale },
+	)
+		.shiftTo("days", "hours", "minutes")
+		.normalize();
+	const unit =
+		duration.days > 0 ? "days" : duration.hours > 0 ? "hours" : "minutes";
```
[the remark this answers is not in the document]
lkeegan: true, refactored into a separate function: f090708

frontend/src/routes/userLand/milestone/+page.svelte
```diff
@@ -79,7 +79,12 @@ async function nextMilestone() {
 		console.log(error);
 		return;
 	}
-	milestoneAnswerSession.answers[`${currentMilestone.id}`] = data;
+	console.log(data);
```
jmsssc, line 82: This could be removed?

mondey_backend/tests/utils/test_statistics.py
```diff
@@ -235,32 +247,34 @@ async def test_calculate_milestone_statistics_by_age(statistics_session, user_se
...
-    await async_update_stats(statistics_session, user_session, incremental_update=False)
-    mg = statistics_session.get(MilestoneGroupAgeScoreCollection, 1)
+    await async_update_stats(session, user_session, incremental_update=False)
+    mg = session.get(MilestoneGroupAgeScoreCollection, 1)
 
     assert mg.milestone_group_id == 1
-    assert mg.scores[8].count == 6
-    assert np.isclose(mg.scores[8].avg_score, (0 + 1 + 1 + 1 + 2 + 0) / 6.0)
-    assert mg.scores[8].stddev_score == pytest.approx(0.719, abs=0.1)
+    assert mg.scores[8].count == len(updated_answers)
```
jmsssc, line 275: Nice this gets around the issue you mentioned with hardcoded tests

mondey_backend/tests/conftest.py
```diff
@@ -1089,7 +1116,7 @@ def active_test_account_user():
 @pytest.fixture
 def active_research_user_full_data_access():
     return UserRead(
-        id=5,
+        id=6,
```
jmsssc, line 1119: This change looks fine but I just wanted to check if there was a reason for it

mondey_backend/src/mondey_backend/statistics.py
```diff
@@ -373,37 +343,60 @@ async def async_update_stats(
...
-            milestone_group_answers[answer.milestone_group_id].append(answer)
+        for milestone in milestones:
+            answer = milestone_answer_session.answers.get(milestone.id)  # type: ignore
+            if answer is not None:
+                milestone_answers[milestone.id].append(answer)  # type: ignore
+                milestone_group_answers[milestone.group_id].append(answer)  # type: ignore
+                logger.debug(f"    - {answer}")
+            else:
+                # if a milestone is not included in the answer session, we still need a value when calculating
+                # the average and std dev of answers for the milestone group.
```
jmsssc, line 365: Because our per-group score needs to be accurate if there is at least one milestone within the group with data right?

as opposed to (wrongly) showing 100% if they only had one milestone (but were much younger than the milestones are usually achieved at) or wrongly showing 25% if only one of four had been recorded?

[based on how you described it yes this code agrees with that approach:
include all milestones in a group when calculating score avg/std dev for that group

    if a milestone was not included in a milestone answer session
        assume a score of 3 if the child is older than the milestone expected age
        assume a score of 0 if the child is younger than the milestone expected age
    this ensures we are comparing avg/std dev across the same populations]

mondey_backend/tests/utils/test_scores.py
```diff
@@ -46,68 +46,71 @@ def test_compute_feedback_simple():
...
-    # answer session 2 scores: [1, 1] -> mean 1.0 -> green
+    # answer session 2 scores for milestone group 1: [1, 1, 0] -> mean 0.66 -> green
     feedback = compute_milestonegroup_feedback_summary(
-        statistics_session, child_id=1, answersession_id=2
+        session, session.get(MilestoneAnswerSession, 2)
     )
     assert feedback[1] == TrafficLight.green.value
     assert len(feedback) == 1
-    for update_existing_statistics in [True, False]:
+    for incremental_update in [True, False]:
```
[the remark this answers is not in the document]
lkeegan: Yes - the variable name was wrong before in the tests - although it didn't affect the correctness particular test since we're just checking that we get the same stats whether this bool is true or false - but now the naming is consistent with how the bool is actually used

mondey_backend/tests/conftest.py
lkeegan: Yup we had two users with the same id (active_test_account_user & active_research_user_full_data_access) - wasn't causing any issues but seems safer to avoid having duplicate ids!

frontend/src/lib/components/DataDisplay/CardDisplay.svelte
```diff
@@ -13,87 +15,111 @@ let {
...
+	if (!data.secondsRemaining) {
+		return "";
+	}
+	const duration = Duration.fromObject(
+		{ seconds: data.secondsRemaining },
+		{ locale: i18n.locale },
+	)
+		.shiftTo("days", "hours", "minutes")
+		.normalize();
+	const unit =
```
jmsssc:
```suggestion
	const unit =
```

We should add translations to this before it goes to production since it just prints the string.

It might be overkill but since we have the users two letter ISO language code we can grab the translation automatically without having to have translations in the admin panel for this one actually, this code is AI generated but gets to what I mean:

function getLocalizedTimeUnit(duration, locale = navigator.language) {
  const rtf = new Intl.RelativeTimeFormat(locale, { numeric: 'auto' });
  
  if (duration.days > 0) {
    // This will return the localized word for "days"
    const formatted = rtf.formatToParts(duration.days, 'day');
    return formatted.find(part => part.type === 'unit').value;
  } else if (duration.hours > 0) {
    const formatted = rtf.formatToParts(duration.hours, 'hour');
    return formatted.find(part => part.type === 'unit').value;
  } else {
    const formatted = rtf.formatToParts(duration.minutes, 'minute');
    return formatted.find(part => part.type === 'unit').value;
  }
}

// Usage examples:
// In Spanish: getLocalizedTimeUnit({days: 2}, 'es') => "días"
// In French: getLocalizedTimeUnit({hours: 3}, 'fr') => "heures"
// In German: getLocalizedTimeUnit({minutes: 30}, 'de') => "Minuten"

mondey_backend/src/mondey_backend/statistics.py
lkeegan: this is needed because for the milestone group traffic light feedback, we compare the child's average score over the milestones for that milestone group with the average & std. dev. for all children of the same age.

this comparison is only valid if each child has provided answers for the same milestones in the group - so if a milestone wasn't asked because the child was too young or too old we need to assume an answer before calculating the averages.

(This avoids problems when the expected ages for a milestone changes, so e.g. a child of the same age would no longer be asked it, or the issue we already have where the children in our datasets have answered all the milestones regardless of expected age, unlike the actual users of the website who will only answers milestones with an expected age within +/- 6 months of the child age)

mondey_backend/src/mondey_backend/routers/utils.py
```diff
@@ -241,22 +263,41 @@ def get_db_child(
...
+    :param count: the count of answers for each age, where the index in the array is the age in months
+    :return: the expected age in months for this milestone
+    """
+    # require at least `min_number_of_answers` answers for each age
+    min_number_of_answers = 3
+    # expected age is the first age with an average score >= `min_avg_score`
+    min_avg_score = 2.1
+    valid_scores = scores.copy()
+    valid_scores[count < min_number_of_answers] = 0
+    return int(np.argmax(valid_scores >= min_avg_score))
```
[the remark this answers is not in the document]
lkeegan: Yup - I guess that is not the best default value, I'll change it to return 72 instead if there is no data or no kids managed the milestone

frontend/src/lib/components/DataDisplay/CardDisplay.svelte
lkeegan: it's actually already localised (at least it works for me locally) - we set the locale when we instantiate the `Duration` - then `.toHuman` prints a localised string

mondey_backend/src/mondey_backend/routers/utils.py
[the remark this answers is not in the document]
lkeegan: 9feddb9b71acfb05b51c564b4d22aa5cc3b43574

mondey_backend/src/mondey_backend/statistics.py
jmsssc: > this is needed because for the milestone group traffic light feedback, we compare the child's average score over the milestones for that milestone group with the average & std. dev. for all children of the same age.
> 
> this comparison is only valid if each child has provided answers for the same milestones in the group - so if a milestone wasn't asked because the child was too young or too old we need to assume an answer before calculating the averages.
> 
> (This avoids problems when the expected ages for a milestone changes, so e.g. a child of the same age would no longer be asked it, or the issue we already have where the children in our datasets have answered all the milestones regardless of expected age, unlike the actual users of the website who will only answers milestones with an expected age within +/- 6 months of the child age)

Ok yep I see that makes sense, thank for explaining here in a comment.

mondey_backend/tests/conftest.py
jmsssc: Definitely! Looks like two changes got merged into the conf test but didn't merge conflict so there was no reason to notice there'd be two IDs unless you happened to look at the whole file excluding changes anyhow when merging in the commit which merged into main later.

jmsssc (approved):
All looks good to me

jmsssc:
@lkeegan though the Gallery background colour and text contrast for the days remaining still should probably be fixed (on light mode this problem is clearer)

PR #364, opened 2025-05-12
Clean up data input

lkeegan:
- child birth year/month are now localised select options
- DataInput no longer overwrites a supplied placeholder
- simplify typing of component type and remove unused types
- remove no longer used / debugging code


frontend/src/routes/userLand/children/registration/+page.svelte
```diff
@@ -156,33 +155,27 @@ async function setup(): Promise<{
...
 }
 
-let birthmonthtext = $derived.by(() => {
-	return birthmonth !== null
-		? Intl.DateTimeFormat(i18n.locale || "en", { month: "long" }).format(
-				new Date(birthmonth.toString()),
-			)
-		: "";
-});
+let birthmonthtext = $derived(
```
jmsssc, line 172: I note that you just need $derived if you are using a ternary in Svelte 5, and $derived.by only when you need a function to do some logic..

lkeegan: yes if it's a one liner then you can just use `$derived` and omit the function definition (although using `$derived.by` with a function is also perfectly fine - as far as I know it's just syntactic sugar)

PR #370, opened 2025-05-14
Clean up gallery display

### LINKED ISSUE #151: integrate styling into GalleryDisplay component data instead of having a separate array for it
This came up in #138  and should be changed to be more robust and less cumbersome to use.

### LINKED ISSUE #365: Fix time remaining text color in child overview
@lkeegan though the Gallery background colour and text contrast for the days remaining still should probably be fixed (on light mode this problem is clearer)

_Originally posted by @jmsssc in https://github.com/ssciwr/mondey/issues/363#issuecomment-2872582593_

lkeegan:
- refactor Gallery arrays of data & style dicts into components with slots
  - add Progress and TimeRemaining components
  - CardDisplay sets the text color based on darkness of card color
  - ChildCardDisplay uses same light/dark text color for time remaining as for other text, resolves #365
  - simplify gallery search bar to only search title text
  - resolves #151 
- improve isDark to check whether black or white provides a higher contrast ratio
  - add tests (some of these test cases failed with the previous isDark implementation)
  - add wcag-contrast-utils to dependencies
- make padding / margins more consistent
- move `/userLand/children/gallery` to `/userLand/children`
- display existing child image on edit child page
- remove no longer used code


jmsssc (changes requested):
Looks good to me and I like refactoring away the component look up table approach, this is great to read. The only one I am not sure with the style changes is the Neu button - the padding works well. Running it local dev on my laptop, fresh with no children, it felt like that text should be white. But it doesn't look wrong amongst others colors, once you add more children.

image

I have requested changes because for now I think we should not display the childrens images (since in a meeting it was requested to remove the child images from the interface for data protection reasons and I believe these still shouldn't be shown nor any functionality to edit images?)

lkeegan:
> Looks good to me and I like refactoring away the component look up table approach, this is great to read. The only one I am not sure with the style changes is the Neu button - the padding works well. Running it local dev on my laptop, fresh with no children, it felt like that text should be white. But it doesn't look wrong amongst others colors, once you add more children.
> 
> image
> 
fixed: [93d140f](https://github.com/ssciwr/mondey/pull/370/commits/93d140fa068d125202f5da58ffa9b7fe4d51b3a6)

> I have requested changes because for now I think we should not display the childrens images (since in a meeting it was requested to remove the child images from the interface for data protection reasons and I believe these still shouldn't be shown nor any functionality to edit images?)

I agree - I thought these features were guarded by the `displayChildImages` bool, or did you see a child image when running it?

jmsssc:
I haven't seen the image on the edit page so no problems visually or anything.

Ok yes if in the future someone was to clone this repo and change this functionality, it makes sense to have the image there. I had removed that code to be on the safe side basically, but yeah it's a bit more consistent that most of the other image code remains to have it remain and preview for children too on the edit page thinking it over.

jmsssc (approved):
All looks good to me

Issue #375, opened 2025-05-16
Z-index issue for logged in menu "Burger" icon


jmsssc:
The menu z-index is not set so the Mondey top fixed bar and logo can(always?) have a higher z-index, so the menu is not accessible. Not sure when this was introduced but the (fairly reliably) fix is to set the z-index of the icon to 1001, one above the z-index 1000 class defaults, rather than the same or no z-index.

Issue #373, opened 2025-05-15
Mobile scrolling issue (causing screen to buzz vertically)


jmsssc:
As described by Igor by email there is a bug sometimes with scrolling. It seems to be

Likely related to this bug: Sometimes on mobile our widths and possibly heights were set beyond 100vh/vw. The Gallery refactor may resolve parts of this by making padding more consistent. However, I have not replicated the bug.

@Majapur1 Please can I ask what device you were using and what app (e.g. Android / Chrome or iOS IPhone / Safari) when the bug appeared? Feel free to email me to keep that info private and I will aim to work out the source of the issue. I have your video too but I haven't attached it here.

Majapur1:
@jmsssc 

Hi James! I was testing on an iPhone 12 Pro Max using Google Chrome.  
I hope this helps!

jmsssc:
So the major issue (Tooltips continously hiding themselves) has been resolved.

I replicated the issue just now with Browser stack on Chrome 12 with iPhone 12 Pro Max.

This is how the phone appears:

![Image](https://github.com/user-attachments/assets/69978c88-7a2c-453e-afe6-f043ccd96526)

jmsssc:
I think it could well be the double layering of 2 m-2s and p-2 with width: 100%:

![Image](https://github.com/user-attachments/assets/cf1315ec-f533-42ca-8722-46cf37743adc)

jmsssc:
(This still needs to be addressed)

jmsssc:
We should confirm this with @Majapur1's phone today during the meeting if possible

Issue #328, opened 2025-04-10
Mobile useability issues


jmsssc:
**Describe the bug**
The mobile interface is too wide for most phones (rather than being 100% width and responsive) so the user scrolls slightly side-to-side. It should be a perfect 100% width.
Long tooltips flash and buzz in and out of the screen on mobile devices, such as when selecting a milestone option like "Gar Nichts".

**Solution**
A) Set widths correctly on mobile. Probably something in the content is width 100% and there is an extra margin or padding.
B) For the tooltips, add a "Help" button below the milestones which has a modal with all the tooltip explanations. We do not use long tooltips anywhere else.

**Device**
Mobile, iOS

**Additional context**
This tooltip will continously re/deselect:

![Image](https://github.com/user-attachments/assets/0b0cb7bf-1438-4815-adb9-2234013fc6bc)

jmsssc:
![Image](https://github.com/user-attachments/assets/e92a7f57-380d-4ec7-8a23-616982ca7130)

Fixed the contrast too.

jmsssc:
Showing the tooltips below does not really work:

![Image](https://github.com/user-attachments/assets/334a9a75-eb2f-4e09-90ac-0cb78f506405)

jmsssc:
I've tried this design:

![Image](https://github.com/user-attachments/assets/ae8d4349-112b-4e96-a8a9-022c45ff0913)

@Majapur1 do you have any thoughts? Would you prefer this (because the current version, the user cannot read the tooltips on some devices)

Majapur1:
@jmsssc I think the last version is perfect! I’ll show it to Mrs. Pauen for her feedback.

Thanks!

lkeegan:
resolved by #376

PR #376, opened 2025-05-16
Small UI fixes - 373, 375 and 328

jmsssc:
This issue fixes a few tiny issues in #373 #375 #328 

The fix in #328 is proposed (awaiting feedback) so I will not merge till we hear back on that.

### DIFF
--- a/frontend/src/lib/components/MilestoneButton.svelte
+++ b/frontend/src/lib/components/MilestoneButton.svelte
@@ -27,13 +27,21 @@ const text_color =
 		: "text-gray dark:text-white";
 </script>
 
-<button
-	type="button"
-	onclick={onClick}
-	class={`${bg_color} ${text_color} ${selected ? 'opacity-100 outline-none ring-4 ring-blue-400' : 'opacity-75 hover:opacity-100'} border-1 m-1 rounded-lg border border-gray-200 px-5 py-3 text-center font-medium md:my- `}
->
-	{@render children?.()}
-</button>
-<Tooltip class={`${bg_color} ${text_color} dark:${bg_color} `}>
-	{tooltip}
-</Tooltip>
+<div class="flex flex-col">
+	<button
+			type="button"
+			onclick={onClick}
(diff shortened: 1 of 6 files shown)


frontend/src/lib/components/MilestoneButton.svelte
```diff
@@ -20,16 +21,27 @@ let {
...
+		{@render children?.()}
+	</button>
+
+	{#if tooltip}
+		<!-- Desktop: Use Tooltip component -->
+		<div class="hidden md:block">
+			<Tooltip class={`${bg_color} ${text_color} dark:${bg_color} `}>
+				{tooltip}
+			</Tooltip>
+		</div>
```
lkeegan, line 45: for me the tooltip also disappeared on desktop - but it seems to work if `hidden md:block` is put inside the Tooltip class instead:
```suggestion
		<Tooltip class={`${bg_color} ${text_color} dark:${bg_color} hidden md:block`}>
			{tooltip}
		</Tooltip>
```

frontend/src/lib/components/MilestoneButton.svelte
```diff
@@ -20,16 +21,27 @@ let {
 } = $props();
 
 let bg_color = `bg-milestone-answer-${index}`;
-const text_color = index < 2 ? "dark:text-black" : "dark:text-white";
+const text_color =
+	index > 2 || index === -1
+		? "text-white dark:text-black"
```
lkeegan, line 26: I find the first couple a bit hard to read in dark mode with white text:
image

lkeegan (approved):
couple small suggestions, otherwise lgtm!

frontend/src/lib/components/MilestoneButton.svelte
jmsssc: Ah that's a really good point and I should check dark mode more often. I will update them all to use dark text on dark mode.

lkeegan:
@jmsssc I'd suggest we merge this so these changes are visible on the deployed website to make it easier to get feedback on them, what do you think?

jmsssc:
I agree so I will merge now we had some feedback too.

PR #377, opened 2025-05-16
Fix the menus on mobile (the top item was not showing) and make the e…

jmsssc:
…mail cut off when it is too long with a tooltip

Which makes it appear like this:
image


jmsssc:
Closed because https://github.com/ssciwr/mondey/pull/379 also includes this commit and these commits did not relate to a specific issue so they can all be reviewed in that PR

Issue #304, opened 2025-04-04
Make PlotLines legend clickable


lkeegan:
The page design (mostly PlotLines) looks smooth now. It makes me wonder if it's possible for variables with many select options for us to have clicking the legend filter out those plot lines? So the user can choose a graph and focus on certain jobs for example:

![image](https://github.com/user-attachments/assets/49fd1d4a-06d8-4005-8ea0-c95c7a3818dc)

_Originally posted by @jmsssc in https://github.com/ssciwr/mondey/pull/300#discussion_r2027537920_

lkeegan:
https://unovis.dev/docs/auxiliary/BulletLegend#interactive-legend

PR #379, opened 2025-05-16
Make plot legends clickable (#304)

jmsssc:
This was completed with the help of AI, basically we make it possible for the user to click each legend item, which toggles whether that data shows on the plot line graph. This also resolves the email length issue on the menus.

### DIFF
--- a/frontend/src/lib/components/DataDisplay/PlotLines.svelte
+++ b/frontend/src/lib/components/DataDisplay/PlotLines.svelte
@@ -6,6 +6,7 @@ import { type PlotData } from "$lib/util";
 import {
 	Axis,
 	BulletLegend,
+	type BulletLegendItemInterface,
 	CurveType,
 	Line,
 	Scale,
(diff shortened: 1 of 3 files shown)


jmsssc:
The commit from #377 has accidentally been included here too. I can remove it but if we merge both, it will merge together fine. So I've closed the other PR to just do it all in this one PR (since both changes are quite small and the other changes did not have assigned issue numbers)

frontend/src/lib/components/DataDisplay/PlotLines.svelte
```diff
@@ -16,22 +17,61 @@ import { onMount } from "svelte";
...
+
+	createChart();
+
+	new BulletLegend(legend_container, {
+		items: legendItems,
+		onLegendItemClick: handleLegendClick,
+	});
+});
+
+function handleLegendClick(item: BulletLegendItemInterface, index: number) {
```
lkeegan, line 42: Do we need to destroy & re-create the chart on every click?

From the example in the docs https://unovis.dev/docs/auxiliary/BulletLegend/#interactive-legend
it looks like we should be able to use the `legend.update` and `chart.updateComponents` methods without having to recreate everything?

lkeegan (changes requested):
thanks - one suggestion on the implementation

frontend/src/lib/components/DataDisplay/PlotLines.svelte
jmsssc: I did try those methods manually but I had no luck.
The problem was that the chart ref got out of date / the on click button handler would be replaced and then not function for the second click depending on how it  was coded.

Using Claude the AI suggested recreation to solve this issue of the onClick handler not working and it doesn't seem to cause any UI interaction. However I also don't like it. I wrote a few light comments around this but I could either investigate or provide a bit more detail, or indeed try to rework this now I have a fresh look at it today, what do you think?

lkeegan: I had a go at this: [7f9a279](https://github.com/ssciwr/mondey/pull/379/commits/7f9a2793622da8cc4abc377634c30c16d2d4b6e0) let me know what you think / if this works for you

lkeegan (approved):

frontend/src/lib/components/DataDisplay/PlotLines.svelte
jmsssc: Yes that works well in testing, sounds like a good solution

PR #381, opened 2025-05-16
Fix bug in stats update

### LINKED ISSUE #378: statistics update fails with database error
**Describe the bug**
After importing more answer sessions, update statistics fails with 

`AssertionError: Dependency rule on column 'milestoneagescorecollection.milestone_id' tried to blank-out primary key column 'milestoneagescore.milestone_id' on instance '<MilestoneAgeScore at 0x7fe8cf91a530>'
`
The same error can be reproduced by reducing the max child age.

The error message seems to be related to cascade deletion, and I wonder if this is due to us inadvertently constructing & deleting objects when we create and then merge these compound objects: 

https://github.com/ssciwr/mondey/blob/4c543d3461bb2a0178ffba1c3803812e56047a8f/mondey_backend/src/mondey_backend/statistics.py#L491-L507

I'll try refactoring the code to simply update the existing stats and see if that makes any difference, as merging can imply creating as well as updating so there might be more going on here than we expect.

lkeegan:
- refactor how existing stats are updated in the database
  - resolves #378
- `analyse_answer_session` now sets `completed=False` for a session which was `completed=True` if it finds an unanswered milestone in the session
  - shouldn't happen for `completed` sessions created by users on the website but may possibly occur with imported data
  - this avoids the possibility of `-1` scores messing up the statistics calculations

### DIFF
--- a/mondey_backend/src/mondey_backend/statistics.py
+++ b/mondey_backend/src/mondey_backend/statistics.py
@@ -226,6 +226,18 @@ def analyse_answer_session(
     diff = 0
     count = 0
     for milestone_id, answer in milestone_answer_session.answers.items():
+        if answer.answer < 0:
+            if milestone_answer_session.completed:
+                logger.warning(
+                    f"    - completed answer session {milestone_answer_session.id} missing score for milestone {milestone_id} - marking session as incomplete!"
+                )
+                milestone_answer_session.completed = False
+                session.commit()
+                session.refresh(milestone_answer_session)
+            logger.debug(
+                f"    - no answer available for milestone {milestone_id} - skipping"
+            )
+            continue
         score = session.exec(
             select(MilestoneAgeScore)
             .where(col(MilestoneAgeScore.age) == child_age)
@@ -236,14 +248,6 @@ def analyse_answer_session(
                 f"    - no statistics available for milestone {milestone_id} - skipping"
             )
             continue
-        if answer.answer < 0:
-            logger.debug(
-                f"    - no answer available for milestone {milestone_id} - skipping"
-            )
-            continue
-        logger.debug(
-            f"    - milestone {milestone_id}: answer {answer}, avg score {score.avg_score}"
-        )
         analysis.answers.append(
             MilestoneAnswerAnalysis(
                 milestone_id=milestone_id,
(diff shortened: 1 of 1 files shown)


mondey_backend/src/mondey_backend/statistics.py
```diff
@@ -226,6 +226,18 @@ def analyse_answer_session(
     diff = 0
     count = 0
     for milestone_id, answer in milestone_answer_session.answers.items():
+        if answer.answer < 0:
```
jmsssc, line 229: Just to check my understanding: This is for a skipped answer (through the UI?) but other code via interface based data entry should protect against ever having such a value if answer session completed = True?

The imported data includes positive values for milestones or absent values as far as I have seen in the data too. I could change the code to assert that - right now indeed it converts the answer value for milestones (which are direct "plaintext" answers, not answer codes) to int and only if they are present/parseable. So I could change that to only save them if they were positive too, just to be on the safe side?

jmsssc (approved):
All looks good to me just one doc string change suggestion.

mondey_backend/src/mondey_backend/statistics.py
lkeegan: > Just to check my understanding: This is for a skipped answer (through the UI?) but other code via interface based data entry should protect against ever having such a value if answer session completed = True?
> 
This is just in case we somehow end up with an answer session marked as completed even though there are missing answers, since including -1 in the avg/std dev calculations would be bad. (I think there are some examples of this in the current data)
It shouldn't happen via the UI, as the `completed` flag is only set when all answers are non-zero.

> The imported data includes positive values for milestones or absent values as far as I have seen in the data too. I could change the code to assert that - right now indeed it converts the answer value for milestones (which are direct "plaintext" answers, not answer codes) to int and only if they are present/parseable. So I could change that to only save them if they were positive too, just to be on the safe side?

I guess we have two options when importing data with `-1` milestone answer values:
1. we only create milestones in the answer session for which we have answers, thus the answer session is `completed` and will be used when calculating statistics
2. we import them as `-1`: the answer session will then get marked as `completed=False` and be ignored when calculating statistics

I think option 1 is probably better, but I guess either way is ok since most of the rows have answers for all milestones (I think?)

jmsssc: > > Just to check my understanding: This is for a skipped answer (through the UI?) but other code via interface based data entry should protect against ever having such a value if answer session completed = True?
> 
> This is just in case we somehow end up with an answer session marked as completed even though there are missing answers, since including -1 in the avg/std dev calculations would be bad. (I think there are some examples of this in the current data) It shouldn't happen via the UI, as the `completed` flag is only set when all answers are non-zero.
> 
> > The imported data includes positive values for milestones or absent values as far as I have seen in the data too. I could change the code to assert that - right now indeed it converts the answer value for milestones (which are direct "plaintext" answers, not answer codes) to int and only if they are present/parseable. So I could change that to only save them if they were positive too, just to be on the safe side?
> 
> I guess we have two options when importing data with `-1` milestone answer values:
> 
>     1. we only create milestones in the answer session for which we have answers, thus the answer session is `completed` and will be used when calculating statistics
> 
>     2. we import them as `-1`: the answer session will then get marked as `completed=False` and be ignored when calculating statistics
> 
> 
> I think option 1 is probably better, but I guess either way is ok since most of the rows have answers for all milestones (I think?)

The data does have empty values for some milestone answers (these get skipped, and no -1 is saved during import I believe). It also has some rows with partially int answers, partially float answers etc which indicates to me some human input has occurred for some of the form data (so for example, the last ca. 106 rows of the full additional data CSV are a good example of this).

Particularly milestones under the SV milestone acronym group have incomplete data. Right now, **in the latest version**, those milestone answers are not saved during import, the code checks for many scenarios to discard and finally only saves if the score is 1 <= {answer} <= 4:
```
![image](https://github.com/user-attachments/assets/acc14772-bc77-4a56-8d92-18f4243eb294)

```

So I think it's a non issue and this is solved basically :+1:

PR #382, opened 2025-05-19
Add child expected milestone age intervals to admin interface

### LINKED ISSUE #366: Add age intervals to admin interface
Currently range of asked milestones is expected age +/- 6 months, regardless of child age.

The range should vary with the child age and be modifiable from the admin interface, e.g. a table like this where an admin user can modify the interval values:

| child age (months) | interval +/- (months)|
| -- | -- |
|0-12 | 6 |
|12-24 | 8 |
|24-36 | 12 |
|36-48 | 18 |
|48-72 | 24 |

The backend would need to store these values in the database and look them up when constructing milestone answer sessions for a child.

lkeegan:
- backend
  - add ChildMilestoneExpectedAgeRange with min/max expected age for each child age to models
    - use these min/max ages when selecting milestones instead of hard-coding child age +/- 6 months
  - add `/child-milestone-expected-age-ranges/` get/post admin endpoints
- admin-frontend
  - add page to edit these age ranges with a simplified interface
  - admin can set the +/- value for a few age ranges, rather than setting a min/max for all 72 ages
  - if desired we can easily modify the frontend to expose more granularity here without any backend changes being needed
- resolves #366

also some minor style changes

- use native range input as the selected value is now invisible for some reason in the flowbite-svelte <Range> component
- remove unnecessary Cards from milestone admin pages


jmsssc:
So firstly I really like the interface, it's clear and works well and I like the text size to indicate what is important!

image

I think the icons being the same is okay too as they are exactly the same thing, and though it felt weird at first, I think them being together next to each other makes sense too as they are
image

**Problem with the Drag'n'drop scales UI?**
For me in dark and light mode the age can be dragged and works but the age dragging UI does not display parts to grab/move or the range (Could be due to my set up?)
image

Is there anything I need to do now to set this up to work and display the sliders @lkeegan ?

https://github.com/user-attachments/assets/06af66c0-372d-4d09-b9b3-69ab1f846c0d

lkeegan:
hmm that's strange - I reverted to using `input type="range"` instead of the flowbite-svelte `Range` component because exactly that part of the UI was broken for me (and for Igor). What the native range input looks like I guess depends on the browser - here is what I see with chrome:

image

(also using the same icon was less a design choice and more just laziness - if you have a suggestion for a better choice let me know)

lkeegan:
maybe this icon?: 
image

jmsssc:
> maybe this icon?: image

Totally agree that would work really well to represent this!

lkeegan:
@jmsssc I tried using a different slider component, could you check if this works for you?

jmsssc:
> @jmsssc I tried using a different slider component, could you check if this works for you?

Sure

jmsssc (approved):
All looks good!
image

lkeegan:
resolves #383

PR #384, opened 2025-05-19
361 suspicious tag should persist when admin sets it

jmsssc:
**Functionality**
All working throughout, inclduing admin panel and tested router changes.

**Tests needed**
Would be best to add a admin_suspicious & admin_not_suspicious answering session with answers that would actually by default put them in the other category and check whether they are correctly in/excluded. Can actually probably do it by some tests manipulating and using the IDs of existing milestone answering sessions.

**Review?**
Not ready, I will mark when ready.


mondey_backend/src/mondey_backend/statistics.py
```diff
@@ -273,17 +274,31 @@ def flag_suspicious_answer_sessions(
...
+        session.exec(
+            select(MilestoneAnswerSession)
+            .where(col(MilestoneAnswerSession.completed))
+            .where(
+                (
+                    col(MilestoneAnswerSession.suspicious_state)
+                    == SuspiciousState.NOT_SUSPICIOUS
+                )
+                | (
+                    col(MilestoneAnswerSession.suspicious_state).is_(None)
```
lkeegan, line 290: `suspicious_state` has `nullable=False` so maybe this line is not needed?

mondey_backend/src/mondey_backend/statistics.py
```diff
@@ -294,7 +309,8 @@ def flag_suspicious_answer_sessions(
                 logger.debug(
                     f"Marking answer session {milestone_answer_session.id} with rms difference {analysis.rms} as suspicious"
                 )
-                milestone_answer_session.suspicious = True
+                # will be properely set now.
```
lkeegan, line 312:
```suggestion
```

mondey_backend/src/mondey_backend/statistics.py
```diff
@@ -332,15 +348,26 @@ async def async_update_stats(
             session.add(answer_session)
         session.commit()
 
-    # get MilestoneAnswerSessions to be used for calculating statistics - exclude any flagged as suspicious or from test accounts
+    # get MilestoneAnswerSessions to be used for calculating statistics
+    # filter to keep only those that are test accounts, or explicitly marked as not suspicious (either by system or admin)
```
lkeegan, line 352:
```suggestion
    # filter to keep only those that are not test accounts and are explicitly marked as not suspicious (either by system or admin)
```

mondey_backend/src/mondey_backend/suspicious_enum_migration.sql
```diff
@@ -0,0 +1,10 @@
+ALTER TABLE milestoneanswersession
+    ADD COLUMN suspicious_state TEXT
+        DEFAULT 'not_suspicious' NOT NULL
+        CHECK (suspicious_state IN ('admin_not_suspicious', 'not_suspicious', 'suspicious', 'admin_suspicious'));
+
+UPDATE milestoneanswersession
+SET suspicious_state = CASE
+                           WHEN suspicious = TRUE THEN 'suspicious'
+                           ELSE 'not_suspicious'
+END;
```
lkeegan, line 10: I guess we should also drop the `suspicious` column here?

lkeegan (approved):
This looks good and seems to work fine with some manual testing, some minor suggestions.

One thing I initially wanted to suggest was to use ints instead of strings for the enum for performance / memory usage reasons, but since this is currently nice and readable and performance doesn't seem to be a problem in general so far I don't think it's worth making this change.

mondey_backend/src/mondey_backend/statistics.py
jmsssc: Yeah I was basically not sure if there was some odd scenario (related to the validators code which isn't strictly needed) where this would be possible. I'll remove it.

mondey_backend/src/mondey_backend/models/milestones.py
```diff
@@ -185,7 +206,29 @@ class MilestoneAnswerSession(SQLModel, table=True):
...
+            return SuspiciousState(v)
+        return v
+
+    # Add model serializer
+    @field_serializer("suspicious_state")
+    def serialize_suspicious_state(self, v):
+        if isinstance(v, SuspiciousState):
+            return v.value
+        return v
+
```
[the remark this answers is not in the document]
jmsssc: Sure. That matches closer to my [initial implementation](https://github.com/ssciwr/mondey/pull/384/commits/9f768fd06b1a50451ee74d4e500c9cdecfff2e84) and intuition too.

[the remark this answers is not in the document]
jmsssc: I've done this so this is ready for re-review tomorrow @lkeegan

PR #389, opened 2025-05-20
Minor refactoring

lkeegan:
- use runes instead of stores for admin stores
- fix some typing issues


jmsssc (approved):
Looks good to me but maybe we can create some kind of single helper for the different runes/store types to all use in adminStore.svelte? Or use a svelte-like version of React-query, like https://tanstack.com/query/v4/docs/framework/svelte/overview in the future?

For example:

```
function createStore<T>(
  fetchFn: () => Promise<{ data?: T[], error?: any }>,
  entityName: string
) {
  let data = $state([] as T[]);
  let isLoading = $state(false);
  let error = $state<string | null>(null);

  return {
    refresh: async (): Promise<void> => {
      isLoading = true;
      error = null;
      
      try {
        const response = await fetchFn();
        if (response.error || response.data === undefined) {
          error = `Failed to get ${entityName}: ${response.error || 'Unknown error'}`;
          data = [];
        } else {
          data = response.data;
        }
      } catch (e) {
        error = `Exception while fetching ${entityName}: ${e instanceof Error ? e.message : String(e)}`;
        data = [];
      } finally {
        isLoading = false;
      }
    },
    get data() {
      return data;
    },
    get isLoading() {
      return isLoading;
    },
    get error() {
      return error;
    },
    clear() {
      data = [];
      error = null;
    }
  };
}
```

lkeegan:
Nice idea, I'll do this now

PR #390, opened 2025-05-20
Improve navigation errors

### LINKED ISSUE #256: Review alerts
Once the website is complete, we should go through usage patterns to identify any alerts / error messages that could be improved, e.g.

- currently get an alert if no data in feedback
  - could just remove this alert, feedback page already says there is no data yet, no need to for an error message
- wrong password on login error message is `Ungültige E-Mail-Adresse oder ungültiges PasswortLOGIN_BAD_CREDENTIALS`
  - the error code should probably go to console logging rather than to the user

lkeegan:
- /children/feedback
  - don't display error if there are no answer sessions yet
- /milestone/* and /children/*
  - redirect to /children if necessary data is missing (e.g. if user refreshed)
- resolves various cases where opening a saved link or refreshing the page results in a broken website state
- resolves #256

### DIFF
--- a/frontend/src/lib/components/DataDisplay/CardDisplay.svelte
+++ b/frontend/src/lib/components/DataDisplay/CardDisplay.svelte
@@ -11,6 +11,7 @@ let {
 	color = undefined,
 	cardClasses = "",
 	titleClasses = "",
+	disabled = false,
 	onclick,
 	children,
 }: {
@@ -21,6 +22,7 @@ let {
 	onclick: () => void;
 	cardClasses?: string;
 	titleClasses?: string;
+	disabled?: boolean;
 	children?: Snippet;
 } = $props();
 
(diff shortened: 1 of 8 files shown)


jmsssc (commented):
✓ don't display error if there are no answer sessions yet

when viewing child:
✓ redirect to /children if necessary data is missing (e.g. if user refreshed) --> http://localhost:5173/userLand/children/registration on refresh opens form, think that's good

✓ resolves various cases where opening a saved link or refreshing the page results in a broken website state:
as far as I looked around, I didn't encounter any unsolved errors.

**???** when recording milestone observations for a child:
"redirect to /children if necessary data is missing (e.g. if user refreshed)": I saw some strange behavior here, not involving refreshing (that worked fine),  where some milestone groups redirected back to the children view - is this intended?)

https://github.com/user-attachments/assets/878ea424-13b1-4305-a25e-0366aace43d9

https://github.com/user-attachments/assets/d77ddb9f-d611-44b6-baa0-cebcbb146a91

I think it's one of the redirects in the two milestone/milestone group file changes that causes this. My suggestion is: If intentional (no relevant milestones?) then we should tell the user rather than just redirect.

[frontend/src/routes/userLand/milestone](https://github.com/ssciwr/mondey/pull/390/files#diff-4a4c6629c20414012ef2b30c36ed9f8c146c9b0d99c73f66fee22ebbd62380a8)

lkeegan:
Good catch! What do you think about disabling milestone group cards that don't contain any milestones instead? There doesn't seem much point in letting the user click on it just to tell them it's empty: https://github.com/ssciwr/mondey/pull/390/commits/921823aa44b02952d39ddb7ed861368bd84f9489

jmsssc:
> Good catch! What do you think about disabling milestone group cards that don't contain any milestones instead? There doesn't seem much point in letting the user click on it just to tell them it's empty: [921823a](https://github.com/ssciwr/mondey/commit/921823aa44b02952d39ddb7ed861368bd84f9489)

Agreed, sure.

jmsssc (approved):
Looks good, redirects as intended.

Issue #386, opened 2025-05-20
Minor Style/Format changes


jmsssc:
![Image](https://github.com/user-attachments/assets/c2624a41-e56e-4837-aa64-9a9f57dbd938)

![Image](https://github.com/user-attachments/assets/b835c4c0-69c9-4b22-8be6-fceef85eded4)

jmsssc:
Because the ProgressBar component only accepts a few hardcoded colors, and our colors are hex codes not CSS classes, there's no quick fix here. We could create our own custom Progress bar component but that's a fair amount of extra (though just plain divs/HTML) code. What do you think @lkeegan ?

![Image](https://github.com/user-attachments/assets/2d095c83-c12e-476f-a1a1-4aee8e3fc10e)

lkeegan:
Looks like we can set it via the `labelInsideClass` prop:

https://flowbite-svelte.com/docs/components/progress#custom-color

Although making it the same color as the background might make it a bit unclear

jmsssc:
> Looks like we can set it via the `labelInsideClass` prop:
> 
> https://flowbite-svelte.com/docs/components/progress#custom-color
> 
> Although making it the same color as the background might make it a bit unclear

We can set the class but the colors are hex codes so unless we limited the colors, we can't just provide classes to it basically, I think.

lkeegan:
What about a class like `bg-[#1da1f2]`? https://v3.tailwindcss.com/docs/customizing-colors#arbitrary-values

jmsssc:
> What about a class like `bg-[#1da1f2]`? https://v3.tailwindcss.com/docs/customizing-colors#arbitrary-values

I tried this and found it did not work in our current set up (which is annoying) but the PR work around worked OK without needing a whole new custom component just for the color/style

PR #392, opened 2025-05-20
Minor style changes

jmsssc:
Completes #386


jmsssc:
Seeing a strange here:
image

I think this is related to how I merged main into this.

jmsssc:
image

frontend/src/lib/components/DataDisplay/Progress.svelte
```diff
@@ -1,19 +1,36 @@
...
-            labelInsideClass={`${progress === 100 ? 'text-white' : 'text-black'} rounded-md text-small pl-2`}
+            labelInsideClass={labelClass}
             progress={`${displayProgress}`}
             size="h-4"
     />
 </div>
+
+<style>
+    /* Define the dynamic background color class that we'll use in labelInsideClass */
+    :global(.dynamic-bg-color) {
```
lkeegan, line 33: just for my understanding, if this is added globally how does this not clash with the other progress bars in the css?

jmsssc: I can't say for certain, but CSS is treated in quite a "automagical" way in svelte I feel. It's not like normal CSS.

What I have noticed is that :global in the context of svelte components seems to be scoped anyway to components below or at the level of the component. I noticed this because I had to use it once before even when it was just used in that one component. I think it acts like a CSS module for the component + anything below it - not universally global like a general .css sheet.

The TLDR I take away is that even when defined as :global, it only applies to things at or below the present component, judging by the phrasing here: https://svelte.dev/docs/svelte/global-styles so far that has aligned with my experience. So since they are all the same level, they don't conflict or define each other and so don't need unique IDs etc.

lkeegan: ok thanks that makes sense

PR #397, opened 2025-05-21
Add translation keys to locales for admins

lkeegan:
- if user is an admin, add [ids] to selectable locales
  - when selected, the id of the text in translations is displayed instead of the translated text
  - helps admins to identify which translation to edit to modify some text on the website
- also some cleanup
  - remove dead code
  - remove console logging
  - use typescript/runes in all components, add missing typedefs
  - inline Frontpage component into /page
  - use `user` store to login and update current user data, remove no longer used utils/login.ts


jmsssc:
@lkeegan  Looks good to me so far however I do notice that the Translations editing admin page now contains a form field for this locale, which when you provide and ID won't show up. We could hide the editing inputs for the [id] locale?

image

lkeegan:
> @lkeegan Looks good to me so far however I do notice that the Translations editing admin page now contains a form field for this locale, which when you provide and ID won't show up. We could hide the editing inputs for the [id] locale?
> 
> image

good catch! fixed in [be605b0](https://github.com/ssciwr/mondey/pull/397/commits/be605b044480f35da0c1c733272e07bdc226a0c8)

jmsssc (approved):
All looks good to me

PR #400, opened 2025-05-21
Improve milestone image sizing

lkeegan:
- use fixed relative width layout on desktop: 1/5 image, 3/5 text, 1/5 buttons
  - resolves issue where image width would change when text changes or beobachtungshinweise was unfolded/folded
- set image width for each breakpoint such that most of the image is not cropped
  - resolves issue where for some screen sizes the image would be very narrow with most of the content cropped


jmsssc (approved):
Nice, looks good

image

PR #401, opened 2025-05-21
Support very basic html formatting in limited translations

jmsssc:
How does this library and approach sound?

Still todo: Stop the tailwind overriding default for the list items (or maybe a workaround like replace the uls/lis in here with a class which restores normal default stylesheet styling for them)


lkeegan:
This looks like a good approach to me. I think adding a class with some `ul` styling locally is probably the way to go - `ul` and `li` are used in various flowbite components so in general they should remain unstyled.

jmsssc:
> This looks like a good approach to me. I think adding a class with some `ul` styling locally is probably the way to go - `ul` and `li` are used in various flowbite components so in general they should remain unstyled.

Done - sure, I only targeted those with the class in a parent container, which are also list related elements

frontend/src/lib/components/FrontPageAccordion.svelte
```diff
@@ -2,17 +2,31 @@
...
+}
 </script>
 
-   <Accordion class="min-w-full px-4 mt-10">
-        <AccordionItem >
+<Accordion class="min-w-full px-4 mt-10">
+    <AccordionItem>
         <span slot="header">{i18n.tr.frontpageAccordion.heading1}</span>
-      <p class="mb-4 mt-4 px-4 text-gray-700 dark:text-gray-400">{@html i18n.tr.frontpageAccordion.summary1}</p>
+        <div class="mb-4 mt-4 px-4 text-gray-700 dark:text-gray-400 whitespace-pre-line restore-default">
```
lkeegan, line 21: If we're rendering html then (at least to me) it's surprising that a newline in the html source is also rendered as a newline - I'd suggest we either do one or the other but not both:
```suggestion
        <div class="mb-4 mt-4 px-4 text-gray-700 dark:text-gray-400 restore-default">
```

lkeegan (approved):
lgtm - one minor suggestion, feel free to change or not as you prefer

frontend/src/lib/components/FrontPageAccordion.svelte
jmsssc: That's true, if the user puts a newline inbetween the bullet points (which I fixed in my translation values), it appears. You're right I'll get rid of that.

PR #403, opened 2025-05-22
Update css names to match use

lkeegan:


jmsssc (approved):
All good! Meant to include this in the previous PR commit as you probably imagined. Was sitting in my local changes rather than in the commit.

PR #406, opened 2025-05-23
Migrate database from sqlite to postgres

### LINKED ISSUE #119: Replace sqlite database with postgres or similar
Sqlite is very convenient for development and testing, but at some point we should migrate to a database that is better suited to be used in production as a website backend, e.g. [postgres](https://www.postgresql.org/)

### LINKED ISSUE #404: nginx 403 for sub-pages when opened directly
**Describe the bug**
Directly opening a sub-page, e.g. https://mondey.lkeegan.dev/userLand/children/ gives a nginx 403

Image

### LINKED ISSUE #405: some completed answer sessions have missing answers
Maybe we should delete all the answer sessions and do a new import of all the data.

But in any case it would make sense to do the existing check for missing answers in "completed" answer sessions for all stats updates, not just on the new sessions for an incremental update

lkeegan:
- backend
  - add database settings:
    - DATABASE_HOST_MONDEYDB
    - DATABASE_HOST_USERDB
    - DATABASE_USER
    - DATABASE_PASSWORD
    - DATABASE_PORT
  - add psycopg to dependencies to support postgres (incl async)
  - if DATABASE_HOST_* is not set a temporary sqlite db is used instead (same behaviour as previously - used for e2e tests)
  - check for "completed" answer sessions with missing answers before doing a stats update
    - resolves #405
- deployment
  - docker-compose.yml
    - add postgres containers for mondey and user databases to docker-compose.yml
    - add restart to backend in case databases are not ready in time
  - fix nginx conf
    - add /index.html to try_files, resolves #404
    - port 80 was accidentally duplicated in https server section
  - update docker image versions and use python 3.13
  - document process used to migrate existing sqlite databases
- delete old files
- resolves #119


docker-compose.yml
```diff
@@ -46,6 +55,38 @@ services:
       options:
         max-size: 20m
         max-file: 25
+  mondeydb:
+    image: postgres
+    restart: always
+    shm_size: 128mb
```
jmsssc, line 61: I'm not sure but could this limit be reached if we import a lot of data or import stats? Is there a reason to have it be limited to 128mb in particular?

PR #410, opened 2025-05-27
Add an age range for each milestone

### LINKED ISSUE #407: add an age range to each milestone
Currently each milestone has an expected age, and there is a +/- for child ages.

So a milestone is given to children with age within the +/- for their age range of the expected age for that milestone.

It would be better to have a +/- value set individually for each milestone.

I.e.

- revert #382 
- add `expected_age_delta` to Milestone, with default value of e.g. 6
- add a slider in the admin interface to set this value
- modify logic to use this when constructing an answer session

lkeegan:
- backend
  - remove `MilestoneChildAgeRange`
  - add `expected_age_delta` to `Milestone`
  - update logic for selecting milestones for an answer session
    - milestone now included if `child_age` within `expected_age_months` +/- `expected_age_delta`
  - update/extend tests
  - add calculated `expected_age_delta` to `MilestoneAgeScoreCollection` with placeholder algorithm
- admin frontend
  - add slider to admin milestone interface for setting `expected_age_delta`
  - move sliders to modal with plot of data
  - restyle PlotScoreAge
  - display expected_age_delta in expected age plot and table
- resolves #407

### DIFF
--- a/frontend/src/lib/components/Admin/MilestoneExpectedAgeModal.svelte
+++ b/frontend/src/lib/components/Admin/MilestoneExpectedAgeModal.svelte
@@ -23,7 +23,7 @@ let scoreCollection = $state(null as MilestoneAgeScoreCollectionPublic | null);
 let expected_age_months = $state(0);
 let expected_age_delta = $state(0);
 
-async function updateAnswerData() {
+async function getScores() {
 	if (!milestone?.id) {
 		return;
 	}
@@ -42,10 +42,10 @@ onMount(async () => {
 		expected_age_months = milestone.expected_age_months;
 		expected_age_delta = milestone.expected_age_delta;
 	}
-	await updateAnswerData();
+	await getScores();
 });
 
-function saveChanges() {
+function applyChangesToMilestone() {
 	open = false;
 	if (milestone) {
 		milestone.expected_age_months = expected_age_months;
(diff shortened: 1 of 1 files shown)


frontend/src/lib/components/Admin/MilestoneExpectedAgeModal.svelte
```diff
@@ -3,27 +3,32 @@
...
-	milestoneId = null,
-}: { open: boolean; milestoneId: number | null } = $props();
+	milestone = $bindable(null),
+}: { open: boolean; milestone: MilestoneAdmin | null } = $props();
 
 let scoreCollection = $state(null as MilestoneAgeScoreCollectionPublic | null);
+let expected_age_months = $state(0);
+let expected_age_delta = $state(0);
 
 async function updateAnswerData() {
```
jmsssc, line 26: The update/saveChanges/get functions here could maybe have different names now we've changed this, to indicate it's just within this CLI and the :bind milestone that gets passed to this modal that gets changed/reloaded to the latest version on load perhaps. I thought at first this component handled the saving but really the parent component handles actually calling the API with update data.

jmsssc: Yep looks great to me

jmsssc (approved):
Looks great

PR #417, opened 2025-05-28
Don't repeat milestones after they have been fully achieved

### LINKED ISSUE #414: Milestones answered with 3 shouldn't be asked again
According to the front page text, if a child has answered 3 for a milestone, then in subsequent sessions this milestone won't be asked again.

backend:
- when constructing a MilestoneAnswerSession, check if there is already a MilestoneAnswer with answer=3
  - if so, then set the answer for the new session to 3, otherwise it should be -1

frontend:
- add a "only show unanswered" checkbox which is checked by default to both milestones and milestone groups pages
  - when checked, any milestones that are already answered (or milestonegroups where all milestones are answered) are hidden
  - when unchecked, all are shown

This should allow us to implement the desired behaviour, while allowing the user to change answers for the pre-filled milestones if desired

lkeegan:
- backend
  - when constructing a new milestone answer session
  - check if there is an existing completed answer session
  - if so, for each milestone, check if that milestone was fully achieved (score=3)
  - in this case, pre-fill the answer to 3 for this milestone
- frontend
  - add a checkbox "only display unanswered" to milestones / milestonegroups
  - when checked, only unanswered milestones / groups with unanswered milestones are displayed
  - when unchecked, all milestones displayed (previous behaviour)
- resolves #414

### DIFF
--- a/DEPLOYMENT.md
+++ b/DEPLOYMENT.md
@@ -170,3 +170,12 @@ Notes on the (one-off) process used to migrate the sqlite databases to postgres 
   - `docker exec -it $(docker ps | grep usersdb | awk '{print $1}') bash`
   - `apk update && apk add pgloader`
   - `pgloader --with "data only" sqlite:///db/users.db pgsql://postgres@127.0.0.1/users`
+
+#### Modifying the database
+
+To modify the database, you can connect to the running postgres database in the `mondeydb` container and then enter an SQL command:
+
+```
+docker exec -it $(docker ps | grep mondeydb-1 | awk '{print $1}') bash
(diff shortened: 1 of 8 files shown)


jmsssc (commented):
I have struggled to get this working: I added the column with 6 as a default value, but it seems like all the milestone groups are disabled to me. I need to look into why this is before continuing to review this PR.

jmsssc:
I  need to look into this more. I think the problem comes that no milestones are considered relevant, and so not returned in the API request, the "milestones" array of the milestones groups is empty. 
image

PR #422, opened 2025-06-02
only show remaining time for session if non-negative

### LINKED ISSUE #421: Timer doesn’t stop after 14 days from the start of the test.
**Issue on the page with kids ./userLand/children**

- The timer doesn’t stop after 14 days from the start of the test and goes negative (in -minutes). 

Image

**machine/OS:**

- OS: iPhone, iOS (Chrome) / macOS (Chrome)

lkeegan:
- resolves #421


jmsssc (approved):
Did exactly this on my local code while checking the problem out. Looks good to me.

PR #426, opened 2025-06-06
Refactor statistics and feedback

### LINKED ISSUE #244: Decide on a cutoff for the standard deviation
Currently, the calculation for the feedback for milestonegroups and milestones works like this: 

- calculate average and standard deviation of the set of answers
- consider score s of child: 
  - red:  s <= avg - 2*stddev 
  - yellow: s <= avg - 1 * stddev 
  - green:  x > avg - 1*stddev 

A problem that can come up is that the standard deviation is very small, such that our average might look like 
3.05 +/- 0.02 or so, i.e., almost everyone answered the same, and only for a few children deviate from this. 

Typically this happens when the sample size is relatively small, and thus it's not representative of the real population. Consequently, the stddev is not a reliable estimator of the true variance. 
If we rely on these calculated numbers in this case, a child that has a score of, say, 2, will be flagged red, even though a score of 2 might be a perfectly reasonable performance in reality (because the real variance is far larger than is captured by our current sample of answers). The consequence is a false alarm. 

In order to avoid this, the current system has a cutoff number for the standard deviation: if it is below 0.01, it does not trust it for the aforementioned reasons, and instead assumes a standard deviation of 1 such that the feedback is calculated like this: 

  - red:  s <= avg - 2
  - yellow: s <= avg - 1 
  - green:  x > avg - 1

Question is, what is a reasonable cutoff number for the standard deviation? is 0.01 ok, or should it be another one? Can it be hard coded or does it have to be specific to certain milestones? 
Secondly, is the assumption of a standard deviation of 1 in these cases ok or should it be another one? 

Related to that, we currently are pessimistic in the feedback calculation. If a child is on the fence (which is only likely to happen when the standard deviation is low and hence not trusted) it is put into the `worse` feedback bracket. Is this correct? 

@Majapur1

### LINKED ISSUE #324: Conservative feedback
If we have less than 20 answers for that age when calculating feedback for a child, we should instead display a grey "not enough data to provide reliable feedback" response instead of red/yellow/green.

For now this can be set to e.g. 5 such that we do actually get some feedback, and later/in production it should be set to 20.

### LINKED ISSUE #409: improve cached stats
For each milestone and child age we currently store 

- avg
- count
- std_dev

This works fine as long as the expected age & feedback can be calculated from these values, which they currently are. (The incremental update of stats is not perfectly accurate but good enough for our purposes)

However for an expected age algorithm like "80% of children of this age score 2 or higher" this is not enough information.

Since there are only four possible answers, we could instead store counts of each answer

- n_0
- n_1
- n_2
- n_3

It is then easy from these counts to calculate the mean & std dev from these 4 values when needed, as well as apply algorithms like the example above.

$N = n_0+n_1+n_2+n_3$
$\langle X \rangle = (n_1 + 2n_2 + 3n_3)/N$
$\langle X^2 \rangle = (n_1 + 4 n_2 + 9n_3)/N$
$\sigma^2 = \langle X^2 \rangle - \langle X \rangle^2$

We could also do the same for the milestonegroups.

Side benefit is that incremental updates are now simpler & don't result in any loss of accuracy.

lkeegan:
- refactor statistics
  - remove incremental stats update to reduce complexity of code / paths to test
    - full update only takes a couple of seconds to run and is done infrequently
  - speed up saving of stats to the database by using a bulk update
  - refactor stats calculation to improve performance
  - calculate average/std dev over milestone groups correctly
- modify milestone feedback to be based on milestone expected age rather than avg/std_dev of scores
  - if child is old enough to be expected to achieve the milestone
    - answer 0 -> red feedback
    - answer 1 -> yellow feedback
    - answer 2 or 3 -> green feedback
  - if the child is too young to be expected to achieve the milestone -> green feedback for all answers
  - this makes the individual milestone feedback independent of the statistics: purely based on expected age which is currently set by the admin
  - resolves #324
- milestone group feedback
  - require at least 5 samples to provide feedback based on avg/stddev comparison
  - resolves #244
- refactor how we cache stats
  - MilestoneAgeScore
    - store answer counts for each answer instead of count/avg/stddev
      - can easily reconstruct count/avg/stddev from these values when needed
      - can also calculate things like "fraction of children answering 2 or 3" - would also allow stacked area plots of answers to be displayed in admin frontend
    - resolves #409
  - MilestoneGroupAgeScore
    - store count, score sum, scoresquared sum
    - easy to do incremental update if needed in the future
- add unknown to SuspiciousState enum
  - new answer sessions intially have this value
  - system then analyses any sessions with this value before stats are calculated
- frontend
  - don't show individual milestone feedback in frontend if not enough samples
  - update plot to match new score data format
  - remove partial stats update button
  - display suspicious state text instead of converting to boolean


lkeegan:
> Looks good to me if you are happy with those two tiny changes I added @lkeegan?

yes - thanks!

jmsssc:
Checking why my .SQL changes broke the test...

lkeegan:
thanks this is much nicer than the previous sql!

PR #428, opened 2025-06-10
Update DEV readme

### LINKED ISSUE #425: Update dev readme
Needs to be updated to work with the postgres setup instead of sqlite.

Probably running postgres locally in a container is the easiest setup?

lkeegan:
- update dev readme on how to run a dev setup locally
- add docker-compose.dev.yml
  - runs only the database images from docker-compose.yml
  - exposes port 5432 for mondeydb and 5433 for usersdb
- backend
  - add separate settings for the two database ports
  - only reload source directory
    - note: uvicorn < 0.34.3 will still add CWD to the folders to scan & reload
  - fix condition to delete temporary sqlite databases
- resolves #425


jmsssc:
> (assuming the permission denied is due to the db folder created by the docker containers?)

Yes exactly, well the reason being that docker is ran as root, so root creates the folders, and your normal user (where I at least have my python environments/run mondey-backend from), cannot then access them. Alternatively, we could advise the reader how to use docker another way or to run mondey on root?

lkeegan:
> > (assuming the permission denied is due to the db folder created by the docker containers?)
> 
> Yes exactly, well the reason being that docker is ran as root, so root creates the folders, and your normal user (where I at least have my python environments/run mondey-backend from), cannot then access them. Alternatively, we could advise the reader how to use docker another way or to run mondey on root?

Did you try upgrading uvicorn? I think the error message is due to uvicorn trying to scan the current working directory to see if any files have changed when running the backend in dev mode, and the latest version stops automatically adding the current working directory to the list of places to scan for source file changes. The mondey-backend shouldn't need to or be able to directly access the folders created by docker (it only interacts with them via the postgres server).

PR #446, opened 2025-08-07
Admin panel option to hide feedback from general users

jmsssc:


lkeegan (approved):
thanks - testing this locally it all works as expected. some minor suggestions, otherwise lgtm!

if in the future we add more settings that affect more frontend pages, it might make sense to refactor these settings into a store that gets fetched in the top level layout.svelte, then the settings can be used in other frontend pages without needing an additional request and logic for each page (i.e. like we do for i18n)

jmsssc:
> thanks - testing this locally it all works as expected. some minor suggestions, otherwise lgtm!
> 
> if in the future we add more settings that affect more frontend pages, it might make sense to refactor these settings into a store that gets fetched in the top level layout.svelte, then the settings can be used in other frontend pages without needing an additional request and logic for each page (i.e. like we do for i18n)

Yeah this sounds good to me to do, if we add more sitewide serverside stored state.

PR #449, opened 2025-08-12
Improve the error message if the user tries to register with an invalid email

### LINKED ISSUE #441: Registration with invalid email leads to broken error message
**Describe the bug**
Upon registering, if you try to register with something like test@localhost (so not a valid email), the error modal is unclear and displays a javascript "[object Object]":

Image

**To Reproduce**
Try to register with an email string which is not a valid email

**Expected behavior**
A generic error saying this is not an email or the account not be created

**Additional context**
The backend returns 422: unprocessable entity, and the frontend fails to handle this well, instead showing [object Object] which is not right.

jmsssc:
Resolves #441

### DIFF
--- a/frontend/src/lib/components/FrontPageAbout.svelte
+++ b/frontend/src/lib/components/FrontPageAbout.svelte
@@ -25,7 +25,7 @@ const registerTestUser = async () => {
 	if (result.error) {
 		alertMessage = `${i18n.tr.registration.alertMessageError}: ${result.error.detail}`;
 		if (result.response?.status === 422) {
-			alertMessage = i18n.tr.login.badData;
+			alertMessage = i18n.tr.login.badCredentials;
 		}
 		showAlert = true;
 	} else {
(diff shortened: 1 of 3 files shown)


jmsssc:
Ready to be reviewed

frontend/src/lib/translations.ts
```diff
@@ -248,6 +248,7 @@ export const translationIds = {
 			"Sie sind bereits angemeldet. Melden sie sich zuerst ab um den Account zu wechseln.",
 		alertMessageTitle: "Fehler",
 		badCredentials: "Ungültige E-Mail-Adresse oder ungültiges Passwort",
```
lkeegan, line 250: Could we use the existing `badCredentials` text here, or does `badData` have a different meaning? (just to avoid duplication if possible)

lkeegan (approved):
one suggestion, otherwise lgtm!

PR #456, opened 2025-08-26
Set min and max values for expected age

### LINKED ISSUE #437: Change expected age + age_delta logic
**Feature request from #438**

- instead of `expected_age` +/- `expected_age_delta`, specify separate lower and upper bounds on relevant ages for a milestone
- a milestone should be included in a session if
  - child age within this range
  - OR the milestone was included in the previous answer session and not fully achieved
- use `expected_age` to provide individual milestone feedback instead of stats
  - e.g. score 0 or 1 with age > expected_age -> "Milestone delayed"
- milestones that are fully achieved should not be asked in future answer sessions
  - we can pre-answer them with 3 when creating the session if in the last session they were present and scored 3
  - frontend should be updated such that after answering a milestone, it jumps to the next **non-answered** question, currently it just goes through all the milestones in that group in the session

**Database changes**

- add a lower and upper bound for ages, e.g.
  - `Milestone.min_relevant_age: int`
  - `Milestone.max_relevant_age: int`
- should probably do the same in `MilestoneAgeScoreCollection`

**Backend changes**

- compute_feedback_simple currently gives stats-based feedback for milestones/milestonegroups
  - add another feedback function for milestones based on age, score & expected_age
- update all uses of Milestone.expected_age_delta and MilestoneAgeScoreCollection.expected_age_delta in code
- when creating a new answer session, need to look up the previous one and
  - add any milestones not present in the new one that got a score < 3
  - pre-answer any milestones that were present in the previous one and scored 3 with a 3
- update logic for calculating milestone expected age / delta accordingly

**Frontend changes**

- after answering a milestone, it should jump to the next **non-answered** question, currently it just goes to the next milestone in the group regardless of it is already answered or not

**Admin frontend changes**

- delta slider -> two text inputs for min/max values
- expected age slider -> text input
- update expected age calculation page

### LINKED ISSUE #438: Adjustments to the specification of "voraussichtliches Alter"
The manual specification of the months for each milestone must contain the following fields:

**1. Anfangszeit:** 

* From which age on (in months) will a given milestone be displayed?

**2. Endzeit:** 

* In case the observation of a given child starts at a later age: which milestones will not be displayed because they are only relevant for younger children?

**IMPORTANT:** 

* Each milestone should be displayed as long as the given child has not yet reached the highest score (zuverlässig). Hence, it may be displayed longer than the “Endzeit” indicates. The Endzeit is only relevant for deciding which milestones will be presented to new children of a given age.
    _(E.g., It makes no sense to evaluate whether 3-year-olds can hold their head stable, but it seems necessary to keep on observing a child on a given milestone that is very delayed.)_

**3. Voraussichtliches Alter:** 

* When is a milestone considered late?

Suggestion: A number field would be better than a slider.

lkeegan:
- NOTE: need to run migration (which deletes the stats), restart backend, then re-generate stats after merging this PR!
- db
  - `Milestone` and `MilestoneAgeScoreCollection`
    - remove `milestone_age_delta`
    - add `relevant_age_min` and `relevant_age_max`
  - update e2e test sql data and conftest data accordingly
- backend
  - get_or_create_current_milestone_answer_session
    - milestones that have previously been fully achieved are no longer included in the answer session
    - however, any unachieved milestones from the previous session are included (even if they are no longer within the relevant age range)
  - statistics update
    - imputed answers for missing milestones are now
      - 0 if the child is below the relevant age (i.e. missing because milestone has never been asked)
      - 3 if the child is old enough for the milestone to be asked (i.e. missing because they have previously achieved it)
- frontend
  - feedback: individual milestones are either ok or delayed
    - if ok, not displayed in milestonegroup dropdown
    - if delayed, displayed in milestonegroup dropdown, but without the previous ?/! icon
- admin frontend
  - use validated text inputs instead of sliders for ages
  - set min/max instead of delta for relevant age range
- resolves #437
- resolves #438


jmsssc (changes requested):
This all looks good to me in general - When creating a new child and adjusting start/end times, the milestones I expect to be selectable appear, depending on what has been answered. I did notice that you can set the minimum year above the maximum year in the UI for editing milestone ages, which should probably not be possible.

lkeegan:
> This all looks good to me in general - When creating a new child and adjusting start/end times, the milestones I expect to be selectable appear, depending on what has been answered. I did notice that you can set the minimum year above the maximum year in the UI for editing milestone ages, which should probably not be possible.

Good point! fixed in 5fe4a353cb15884dab992d7d781c1c36bedadd97

PR #460, opened 2025-08-27
Add automatic database migration

### LINKED ISSUE #454: automatic database migrations
Currently if a PR modifies a database model, the docker containers are automatically updated, but the database migration needs to be **manually** done on the deployed database. This is ok for now, but once we are no longer hosting the website this will be become an issue.

A better solution would be to use a database migration tool like alembic, and to do the migration during/before the fastAPI app startup. This would allow us to merge a PR with database model changes (and the corresponding alembic changes) - then when watchtower pulls the latest docker images the migration would take place automatically on app startup.

**Describe alternatives you've considered**
The alembic migration step could be done as a separate command before launching the backend in the backend Dockerfile to avoid having to integrate this with the fastAPI app.

lkeegan:
- add alembic to backend dependencies
- customise alembic/env.py
  - import our models
  - use our database engine (reads `DATABASE_PASSWORD` from the environment the same as mondey-backend)
  - exclude F401 ruff rule for this file (to avoid ruff from removing the 'unused' model imports)
- customise alembic/script.py.mako
  - add sqlmodel import
- backend docker image runs migrations before starting backend
- add README with info on how to run and create migrations as a developer
- update FAQ accordingly
- remove all previous .sql migration files
- resolves #454


jmsssc:
@lkeegan is there any chance you have a `alembic.ini` file locally (or globally?) that should be part of this PR?

PR #470, opened 2025-09-29
Fix bug where missing translation modal would skip translations

### LINKED ISSUE #469: missing translations modal skips every other missing text
**Describe the bug**
After clicking save, the next missing text is skipped - this is because we increment the current index of the missing translations array as well as updating this array (which removes the just-saved text from the start of it) so we effectively skip ahead by 2 every time

lkeegan:
- previously idx would be incremented *and* the missing translations array would be updated
- this means every other missing translation was skipped
- simplify logic to just always displays the first missing translation, since on save this array gets updated
- resolves #469


jmsssc:
> @jmsssc thanks this looks good to me - but I can't approve it since I originally opened the PR, so feel free to approve & merge :-)

Yes sorry, I didn't have time on Friday to run this locally, I was also working on the ENV/port change which just dragged on when I was consolidating it. That's why I hadn't approved yet [I had it in mind throughout the week].

lkeegan:
great, thanks for testing this!

PR #474, opened 2025-10-06
Use dynamic env variables for all ports

jmsssc:
This moves the backend ENV variables to the root directory and provides sample .env files for development. It allows all ports (database, frontend, and backend) to be configured.

To avoid possible leaking of secrets, frontend and backend env variables are still split into two different files. But since frontend only loads VITE_ variables as you can see in /frontend/vite.config.ts, we could combine the two .env files into just one (and that way also only define ports once, e.g. Backend can still access VITE variables so ports/urls could be defined with VITE_, but frontend could only access the VITE ones).

To merge you need to update the .env in both / and /frontend and it would make sense to delete any .env you have in src/mondey_backend to avoid confusion.

The values, in this PR, should be the same as the previous defaults, so running local dev should not change.

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -27,14 +27,7 @@ cd mondey
 openssl req -x509 -newkey rsa:4096 -keyout key.pem -out cert.pem -sha256 -days 365 -nodes -subj '/CN=localhost'
 ```
 
-3. define settings in a .env file in the same folder as the `docker-compose.yml` file, eg:
-
-```
-SMTP_HOST=""
-SECRET="abc123"
-DATABASE_PASSWORD="abc123"
-LOG_LEVEL="debug"
-```
+3. define settings in a .env file in the same folder as the `docker-compose.yml` file. A sample .env.sample file is included in the root directory, which can be copied to .env, with the relevant VALUEs filled in(SECRETS/API keys)
 
 3. build and run the website locally in docker containers on your computer:
 
--- a/.env.sample
+++ b/.env.sample
@@ -1,0 +1,35 @@
+# Change the SECRETVALU values to random alphanumeric strings. For APIs, like Deepgram, request a value, and then change APIVALUE it. Then, after all VALUEs are filled out, copy this to .env
+
+# Database configuration
+DATABASE_PASSWORD=SECRETVALUE
+DATABASE_HOST_MONDEYDB=localhost
+DATABASE_HOST_USERSDB=localhost
+DATABASE_USER=postgres
+DATABASE_PORT_MONDEYDB=5432
+DATABASE_PORT_USERSDB=5433
+
+# Backend server configuration
+SECRET=SECRETVALUE
+HOST=localhost
+PORT=8000
+SMTP_HOST=""
+LOG_LEVEL="debug"
+RELOAD=true
+COOKIE_SECURE=false
+STATS_CRONTAB="0 3 * * mon"
+
+# External services
+DEEPL_API_KEY=APIVALUE
+# Derived from VITE_HOST:VITE_PORT in /frontend/.env(.sample) for backend use - would be better if automated but this is okay
+MONDEY_HOST=localhost:5173
+
+# File paths
+STATIC_FILES_PATH=static
+PRIVATE_FILES_PATH=private
+
+# Testing
+E2E_TEST_USER_SQL_FILES=""
+E2E_TEST_MONDEY_SQL_FILES=""
+
+# Application settings
+MAX_CHILD_AGE_MONTHS=72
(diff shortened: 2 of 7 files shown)


.github/workflows/ci.yml
```diff
@@ -116,6 +116,8 @@ jobs:
         run: pnpm install
       - name: Build frontend
         working-directory: ./frontend
+        env:
```
lkeegan: since this is building the docker frontend image I don't think we need to set VITE_MONDEY_API_URL here

DEVELOPMENT.md
```diff
@@ -27,14 +27,7 @@ cd mondey
...
 
-3. define settings in a .env file in the same folder as the `docker-compose.yml` file, eg:
-
-```
-SMTP_HOST=""
-SECRET="abc123"
-DATABASE_PASSWORD="abc123"
-LOG_LEVEL="debug"
-```
+3. define settings in a .env file in the same folder as the `docker-compose.yml` file. A sample .env.sample file is included in the root directory, which can be copied to .env, with the relevant VALUEs filled in(SECRETS/API keys)
```
lkeegan: I don't think we have a root .env.sample? And since the defaults in the docker-compose are (hopefully) sensible values for deployment I'd suggest we only put the values that need to be changed in .env.sample, i.e. SECRET, DATABASE_PASSWORD, DEEPL_API_KEY, MONDEY_SSL_CERT, MONDEY_SSL_KEY, RELAYHOST

lkeegan (approved):
thanks - two minor suggestions otherwise lgtm!

frontend/src/routes/+layout.ts
```diff
@@ -1,6 +1,9 @@
 import { client } from "$lib/client/client.gen";
 export const prerender = true;
 
+const apiUrl = import.meta.env.VITE_MONDEY_API_URL;
+console.log("Mondey API URL configured:", apiUrl);
```
lkeegan: remove the console.log?

DEVELOPMENT.md
jmsssc: > ok sounds good - depending on how many extra ones are needed for local docker dev it might be worth having a .env.prod.sample and .env.dev.sample?

I think, in practise, only MONDEY_HOST is different in prod (as it was before) (as you can get errors otherwise or functionality doesn't work, e.g. when lacking an API key).

I can see the value, as having mondey.de as the settings.py default (so for prod) but having the sample use the local dev location is strange, it means you need to ignore the .env.sample value and leave the production .env actually empty (so it uses the settings.py default) or have it present twice (in the .env file). I'll do that since it's quite small and then is quite nice and clear what changes between production and development too.

frontend/.env.prod.sample
```diff
@@ -0,0 +1,4 @@
+VITE_HOST=localhost # not sure if this is mondey.de in prod?
+VITE_PORT=5173 # not sure if this is 443/80 in prod (or we use nginx or something to direct traffic on 443 to port 5173?)
+VITE_API_PROXY_URL=http://localhost:8000 # I think this is still local host in production because we are directing these from the frontend containers vite proxy running on the machine through the network docker compose has to the backend
+VITE_MONDEY_API_URL=https://mondey.de/api # but this needs to reference the production site since this is where the vite proxy server is running.
```
lkeegan: We don't need a frontend .env for production - `VITE_MONDEY_API_URL` is hard-coded to `/api` in the docker-compose, and the rest of these variables are only used by the vite dev server

jmsssc:
@lkeegan I think this resolves the issue on both this branch and admin-documents. I struggle to pin down where this came from, but this async update code (which I last updated about 4 months ago when adding some more e2e tests) has been causing the root error of the e2e tests going down (a case in point of what you said today where an E2E test failing could be anything down the pyramid, like here, the backend!) So please can I get your go ahead on this commits fix, will this be ok in production?:

https://github.com/ssciwr/mondey/pull/474/commits/64688d1b99f164b6ccfcb7db9b6d9f3aeafff4af

lkeegan:
> @lkeegan I think this resolves the issue on both this branch and admin-documents. I struggle to pin down where this came from, but this async update code (which I last updated about 4 months ago when adding some more e2e tests) has been causing the root error of the e2e tests going down (a case in point of what you said today where an E2E test failing could be anything down the pyramid, like here, the backend!) So please can I get your go ahead on this commits fix, will this be ok in production?:
> 
> [64688d1](https://github.com/ssciwr/mondey/commit/64688d1b99f164b6ccfcb7db9b6d9f3aeafff4af)

Thanks for looking into this - your fix looks fine to me. (I guess we might have ran into https://github.com/JasperSui/fastapi-injectable/issues/185)

PR #511, opened 2026-07-16
Fix silent credential expiry

### LINKED ISSUE #503: credentials silently expire
Apparently user login has a timeout, and after this while everything seems to still be working, clicking Save doesn't do anything, and when the user finally refreshes they realise they are logged out and have lost their changes.

We should show a warning when the user is close to being logged out, asking them to do something to avoid being logged out.

Also refresh login on every API call, if not already done?

And if the logout does happen this should be instantly visible to the user, should not be left in what looks like working logged in state when actually they are not logged in and API calls are not working.

First step is to reproduce the issue, probably the timeout can be set to e.g. 1min to do this

lkeegan:
- new auth invalidation scheme
  - require some authenticated API request every `SESSION_IDLE_TIMEOUT_SECONDS` (default 1hr)
  - require password every `SESSION_ABSOLUTE_TIMEOUT_SECONDS` (default 8hr)
  - warn `SESSION_WARNING_SECONDS` before either of the above occur (default 5min)
  - update timestamp at most every `SESSION_TOUCH_INTERVAL_SECONDS` (default 5min)
- backend
  - existing `accesstoken.created_at` determines absolute expiry
  - add `accesstoken.last_seen_at` to determine idle expiry
  - extend alembic migration code to users db as well as mondey db
  - /auth/session/refresh — resets last_seen_at which extends the idle timeout
  - /auth/session/reauthenticate — verifies the current user’s password, rotates the access token and resets created_at and last_seen_at which extends both timeouts
- frontend
  - display message before timeout due to inactivity
    - clicking continue or doing anything else involving an API call resets the timeout
    - if user does nothing, redirects to login page at expiry
  - display message asking for password before session timeout
    - user enters password, gets new auth session
    - if user does nothing, redirects to login page at expiry
  - global handler for 401
    - displays message, then redirects to login page
- resolves #503


lkeegan:
On approaching idle timeout:
image

On approaching absolute timeout:
image

